Issue #48, opened 2018-06-21
Configure webpack to minify src on production


dapplion:

dapplion:
After testing it I managed to reduce the size of the package by ~10%, which is not a relevant reduction (4-5Mb) against an ISO of ~800Mb. Will revisit this in the case that we change to a lighter linux distro.

PR #106, opened 2018-10-06
Reduced image size

dapplion:


dapplion:
On hold:
https://github.com/dappnode/DNP_DAPPMANAGER/issues/48

> After testing it I managed to reduce the size of the package by ~10%, which is not a relevant reduction (4-5Mb) against an ISO of ~800Mb. Will revisit this in the case that we change to a lighter linux distro.

PR #158, opened 2019-01-21
added second threshold + improved messages

dapplion:


dapplion:
@Shelpin Can you verify that the thresholds work as expected? Thank you!

PR #200, opened 2019-05-15
Fix bitcoin url typo

dapplion:


dapplion:
@Shelpin Can you test this change please, and make sure the bitcoin RPC is able to connect?

PR #204, opened 2019-05-17
dont load DNP until all downloads are successful

dapplion:
If you install DNP_A which depends on DNP_B, it is very possible that DNP_A download succeeds but DNP_B doesn't because it's not available or it hasn't propagated yet. If that happened, DNP_A will be loaded and it's manifest and docker-compose will be changed. But the install will error.

To fix this dangerous intermediate state, now the installation is split in 3 steps:
1. Download: download all of DNP's images. If a single image is not available the installation will stop at this stage.
2. Load: load the docker image and overwrite the manifest and docker-compose of each DNP.
3. Run: docker-compose up each DNP

### DIFF
--- a/build/src/src/calls/installPackage.js
+++ b/build/src/src/calls/installPackage.js
@@ -130,15 +130,20 @@ const installPackage = async ({
       return { name, ver, manifest };
     })
   );
-  logs.debug(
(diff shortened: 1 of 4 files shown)


dapplion:
@eduadiez suggestion:
> on `packages.load`, load image FIRST, THEN write the manifest and docker-compose

PR #216, opened 2019-05-26
Add --force-recreate flag to restartPatch container

dapplion:


dapplion:
Tested and works fine

PR #219, opened 2019-06-07
Limit from transfers size

dapplion:
Limit directory size before and after compression,
- 200MB precompresion
- 20MB compressed

Also, compress as .zip


dapplion:
Tested successfully with `fallocate -l 24M filename`

PR #237, opened 2019-07-08
Change host user password calls

dapplion:


build/src/src/calls/sshPasswordIsSecure.js
```diff
@@ -0,0 +1,18 @@
+const { isSshPasswordSecure } = require("modules/sshPassword");
+
+/**
+ * Changes the SSH password of the host for the user `dappnode`
```
vdo: This is not the SSH password, is the system wide user password, so the naming is not the best.

build/src/src/modules/sshPassword.js
```diff
@@ -0,0 +1,86 @@
...
+const insecureSalt = "insecur3";
+
+/**
+ * The password is set in the pre-seed, only affects ISO-dappnodes
+ * https://github.com/dappnode/DAppNode_Installer/blob/472d046f2661d737c4650a1f895c8c4d89234cd8/build/dappnode/scripts/preseed_unattended.cfg#L12
+ * `d-i passwd/user-password-crypted password $6$insecur3$rnEv9Amdjn3ctXxPYOlzj/cwvLT43GjWzkPECIHNqd8Vvza5bMG8QqMwEIBKYqnj609D.4ngi4qlmt29dLE.71`
+ *
+ * The file `/etc/shadow` contains:
+ * - with the original salt and password
+ * `dappnode:$6$insecur3$rnEv9Amdjn3ctXxPYOlzj/cwvLT43GjWzkPECIHNqd8Vvza5bMG8QqMwEIBKYqnj609D.4ngi4qlmt29dLE.71:18004:0:99999:7:::`
```
vdo: I think we are giving too much detail here, that can be exploited, so I would remove all those comments

build/src/src/modules/sshPassword.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @param {string} newPassword super-secure-password
+ */
+async function changeSshPassword(newPassword) {
+  if (!newPassword) throw Error("newPassword must be defined");
+
+  if (await isSshPasswordSecure())
+    throw Error(
+      "The SSH password can only be changed if it's the insecure default"
+    );
```
vdo: I think we should return an error if the password is shorter than 8 chars

build/src/src/calls/sshPasswordIsSecure.js
dapplion: Ok, renaming all functions and variables

build/src/src/modules/sshPassword.js
dapplion: Ok, deleted comments

build/src/src/modules/sshPassword.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const image = await getDappmanagerImage();
+  const res = await shell(
+    `${baseCommand} ${image} sh -c "echo dappnode:${newPassword} | chpasswd"`
+  );
+
+  // Check the return for success? #### TODO
+  return res;
```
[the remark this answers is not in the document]
dapplion: @vdo Replies that it suffices to check the exit status

build/src/src/modules/sshPassword.js
dapplion: Ok

dapplion:
Tested on a DAppNode by installing the branches `change_ssh_password` of both the DAPPMANAGER and the ADMIN.
It successfully detected the insecure password, performed the change, and then detected that it was no longer insecure.

PR #255, opened 2019-08-30
Use side container for power

vdo:
lint


build/src/src/calls/installPackage.js
```diff
@@ -181,7 +181,7 @@ const installPackage = async ({
     //   lockedPortsToOpen = [ {portNumber: '32769', protocol: 'UDP'}, ... ]
     await lockPorts(pkg.name || pkg.manifest.name);
   }
-  
+
```
dapplion: Critical change, much needed, thanks <3 !!

PR #304, opened 2019-12-17
Detect usb symlinks

vdo:


build/src/hostScripts/detect_fs.sh
```diff
@@ -12,12 +12,22 @@ for line in $(df -hlPT | grep ^/); do
     free=$(echo "$line" | awk '{ print $5 }')
     partition=$(echo "$line" | awk '{ print $1 }' | xargs basename)
     device="$(readlink -f "/sys/class/block/$partition/.." | xargs basename || true)"
-    model="$(lsblk -no MODEL /dev/$device 2> /dev/null | sed -e 's/[[:space:]]*$//' || true)"
-    vendor="$(lsblk -no VENDOR /dev/$device 2> /dev/null | sed -e 's/[[:space:]]*$//' || true)"
-	  # print in JSON format
-	  [[ "$NEXT" != "true" ]] || echo -n ","
+    model="$(lsblk -no MODEL /dev/$device 2>/dev/null | sed -e 's/[[:space:]]*$//' || true)"
```
eduadiez: Should we use the same algorithm that usbmount to detect the model and version?
```
# Determine vendor and model.
vendor=
if [ -r "/sys$DEVPATH/device/vendor" ]; then
    vendor="`cat \"/sys$DEVPATH/device/vendor\"`"
elif [ -r "/sys$DEVPATH/../device/vendor" ]; then
    vendor="`cat \"/sys$DEVPATH/../device/vendor\"`"
elif [ -r "/sys$DEVPATH/device/../manufacturer" ]; then
    vendor="`cat \"/sys$DEVPATH/device/../manufacturer\"`"
elif [ -r "/sys$DEVPATH/../device/../manufacturer" ]; then
    vendor="`cat \"/sys$DEVPATH/../device/../manufacturer\"`"
fi
vendor="$(echo "$vendor" | sed 's/^[[:blank:]]\+//; s/[[:blank:]]\+$//')"

model=
if [ -r "/sys$DEVPATH/device/model" ]; then
    model="`cat \"/sys$DEVPATH/device/model\"`"
elif [ -r "/sys$DEVPATH/../device/model" ]; then
    model="`cat \"/sys$DEVPATH/../device/model\"`"
elif [ -r "/sys$DEVPATH/device/../product" ]; then
    model="`cat \"/sys$DEVPATH/device/../product\"`"
elif [ -r "/sys$DEVPATH/../device/../product" ]; then
    model="`cat \"/sys$DEVPATH/../device/../product\"`"
fi
model="$(echo "$model" | sed 's/^[[:blank:]]\+//; s/[[:blank:]]\+$//')"
```

vdo: I think our solution is simpler and works for all cases until now, so if we don't see any reason to change I would keep it like it is now. We can adopt it in the future if we feel like it's a better approach.

PR #357, opened 2020-05-05
BIND issue patch

dapplion:
- Remove new legacy bind volume
- When the BIND is re-created, run `nsupdate` on all domains

### DIFF
--- a/build/src/src/modules/legacy/removeLegacyBindVolume.ts
+++ b/build/src/src/modules/legacy/removeLegacyBindVolume.ts
@@ -10,25 +10,24 @@ const newBindVol = "dncore_binddnpdappnodeeth_bind";
 /**
  * After core version 0.2.30 there will be an orphan volume of the
  * DNP_BIND which should be removed for asthetic versions
+ *
+ * After core version 0.2.34 the new volume will become orphan and
+ * must be removed for the bind state to be reset
  */
 export async function removeLegacyBindVolume(): Promise<void> {
   const volumes = await dockerVolumesList();
   const volumeExists = (name: string): boolean =>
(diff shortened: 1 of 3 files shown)


build/src/src/modules/legacy/removeLegacyBindVolume.ts
```diff
@@ -10,25 +10,24 @@ const newBindVol = "dncore_binddnpdappnodeeth_bind";
...
-    if (usersOfOld)
-      throw Error(`legacy BIND volume ${oldBindVol} has users: ${usersOfOld}`);
-
-    // Delete only if has no users
-    await dockerVolumeRm(oldBindVol);
-    logs.info(`Removed legacy BIND volume ${oldBindVol}`);
+      // Delete only if has no users
+      await dockerVolumeRm(oldBindVol);
+      logs.info(`Removed legacy BIND volume ${oldBindVol}`);
+    }
```
dapplion, line 31: Generalize the existing removeLegacyBindVolume to kill both BIND volumes if the exist and have no users

build/src/src/watchers/nsupdate/index.ts
```diff
@@ -70,9 +70,11 @@ setIntervalDynamic(() => {
   nsupdateInterval //      60 min
 ]);
 
-// React immediatelly to new installs
 eventBus.packagesModified.on(({ ids, removed }) => {
```
dapplion, line 73: Event `packagesModified` is called when a container is recreated, on `updatePortMappings`, `updatePackageEnv`, `restartPackageVolumes`, `restartPackage`, `removePackage`, `installPackage`, `backupRestore`

PR #423, opened 2020-10-30
Pablo docker timeout

pablomendezroyo:
**Added timeout parameter to:**
- `dockerComposeUp():` called in restartPackage, restartPackageVolumes, rollbackPackages, runPackages, and tests.
- `dockerComposeDown():` called in packageRemove and tests.
- `dockerComposeStop():` no needed to change anything in the background since it is not called anywhere.
- `dockerStop():` called in packageStartStop and tests.

**Added new funtion:**
- `getDockerTimeoutMax():` this function returns the maximum timeout value from all the docker packages, in order to be used in the corersponding place.

**Other changes**
These changes were implemented after running yarn build
- Due to the delete of the timeout parameter in some functions in the previous commit, there have appeared some errors due to functions calls with this parameter. There have been deleted the timeout parameter in these function calls.
- Added timeout parameter to coreUpdatePackageData()
- Give undefined value to timeout parameter in getInstallerPackageData()

### DIFF
--- a/packages/dappmanager/src/modules/docker/restartPackageVolumes.ts
+++ b/packages/dappmanager/src/modules/docker/restartPackageVolumes.ts
@@ -1,7 +1,7 @@
 import fs from "fs";
 import { uniq } from "lodash";
 import { dockerRm, dockerComposeUp } from "./dockerCommands";
-import { listPackage, listPackages } from "./listContainers";
+import { listPackages } from "./listContainers";
 import { removeNamedVolume } from "./removeNamedVolume";
 import params from "../../params";
 // Utils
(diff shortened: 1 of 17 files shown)


pablomendezroyo:
no more tests errors found

packages/dappmanager/src/modules/docker/utils.ts
```diff
@@ -1,3 +1,5 @@
+import { PackageContainer } from "../../common";
```
dapplion, line 1: PLease import from types not common

packages/dappmanager/src/modules/docker/restartPackageVolumes.ts
```diff
@@ -1,7 +1,7 @@
 import fs from "fs";
 import { uniq } from "lodash";
 import { dockerRm, dockerComposeUp } from "./dockerCommands";
-import { listPackages } from "./listContainers";
+import { listPackage, listPackages } from "./listContainers";
```
dapplion, line 4: If listPackage is not used please remove it

dapplion (changes requested):

dapplion (approved):

PR #424, opened 2020-11-03
Added parseTimeout function and unit test for that function

pablomendezroyo:
**Changes:**
- Added `parseTimeout()` function.
- Called `parseTimeout()` function where needed (getInstallerPackageData)
- New unit test for testing the function.
- Added new functionality to `parseTimeout()` function: returns undefined when the parameter is undefined.

### DIFF
--- a/packages/dappmanager/src/utils/timeout.ts
+++ b/packages/dappmanager/src/utils/timeout.ts
@@ -1,0 +1,31 @@
+// timestring does not have a @types package
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import timestring from "timestring";
+
+// Define build timeout (60 min)
+const defaultBuildTimeout = 60 * 60 * 1000;
+
+/**
+ * Parses a timeout string and returns a number in miliseconds
+ * @param timeout "20min", "5000", undefined
+ */
+export function parseTimeout(timeout: string | undefined): number | undefined {
+  switch (typeof timeout) {
+    case "number": {
+      return timeout;
+    }
+    case "string": {
+      if (!timeout) defaultBuildTimeout;
+      // Timestring returns in seconds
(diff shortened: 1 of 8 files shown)


packages/dappmanager/src/utils/timeout.ts
```diff
@@ -0,0 +1,31 @@
+// timestring does not have a @types package
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
+import timestring from "timestring";
+
+// Define build timeout (60 min)
+const defaultBuildTimeout = 60 * 60 * 1000;
```
dapplion, line 8: This default build timeout is for the SDK only. parseTimeout should return undefined in the default switch case

packages/dappmanager/test/utils/timeout.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { parseTimeout } from "../../src/utils/timeout";
+
+describe("Util: timeout", () => {
+  it("It should parse string parameter to number", () => {
+    const str = "20min"; // 12000 segs, 1200000ms
+    expect(parseTimeout(str)).to.equal(1200000);
+  });
+  it("It should return undefined timeout when the given parameter is undefined", () => {
+    expect(parseTimeout(undefined)).to.equal(undefined);
+  });
```
dapplion, line 13: Add a test for the number case

dapplion (changes requested):

packages/dappmanager/test/utils/timeout.test.ts
pablomendezroyo: Done. Anyway, I though that since the function only accepted `string | undefined `parameters, the function would not be never called with number as argument

dapplion: Right but when dealing with external uncontrolled inputs it's best to cover all cases

dapplion (approved):

PR #425, opened 2020-11-10
Split host stats API endpoints

### LINKED ISSUE #422: Show host stats values in UI dashboard
Right now, in the UI only the % of taken disk space is shown. Should also the numeric values, e.g. 254GB/500GB

- [x] Split DAPPMANAGER `calls/getStats` function into separate functions for cpu, memory and disk
- [x] Edit the current functions to also send absolute data as well as a ratio
- [x] Modify API schema to reflect the new type
- [x] Edit the UI to also include the absolute data formated with `prettyBytes`

#### Memory

```
lion@lion$ free
              total        used        free      shared  buff/cache   available
Mem:        7903464     6315364      164472      692516     1423628      711632
Swap:       8117756     2197916     5919840
```

```ts
interface HostStatMemory {
  memTotal: number;
  memUsed: number;
  used: number; // = Swap.used / Mem.total
}
```

#### Disk
 
```
lion@lion$ df /
Filesystem                1K-blocks      Used Available Use% Mounted on
/dev/mapper/mint--vg-root 235782040 176187744  47547552  79% /
```
```ts
interface HostStatDisk {
  used: number;
  available: number;
  used: number;
}
```

#### CPU

```
?
```

```ts
interface HostStatCpu {
  used: number;
}
```

------

Current UI

Screenshot from 2020-10-29 18-34-04

-------

Target result

Screenshot from 2020-10-29 18-34-04 (1)

pablomendezroyo:
Fixes #422: Show host stats values in UI dashboard

**Main changes:**

1. Edited commands to retrieve host machine stats. Added flags to retrieve stats in bytes and be more consistent.

- `df / ` ->  `df / --block-size=1`
- `free /`  -> ` free / --bytes`

2. Creation of 3 new interfaces in types for memory (HostStatMemory), cpu (HostStatCpu) and disk(HostStatDisk). 

- **HostStatMemory:** have as arguments the output from the command `free / --bytes`.
- **HostStatDisk:** have as arguments the output from the command `df / --block-size=1`.
- **HostStatCpu:** idem as before.
* In order to be consistent with the data, all the percentages will be without the symbol '%', and that symbol will be added in the UI.

3. Splitted getStats() function into three: 

- **statsMemoryGet():** added utility function to parse the command output.
- **statsDiskGet():** added utility function to parse the command output.
- **statsCpuGet()**: no changes

3. Changes in HostStats.tsx file:

- **Added DiskStats JSX element:** will print out the diskUsed(GB) / diskTotal(GB)
- **Loading element:** when the api cannot fetch the data because there is an error, there will be shown a Loading element with error message in the screen. When the api cannot fetch the data but there is not an error, there will be shown loading element with loading message. This element has been reused from the components folder.

4. **Added unit tests:** only added for statsDiskGet() and statsMemoryGet(). 

Other changes:

1. **CSS** for DiskStats JSX element: css style added when printing out the disk usage vs disk total.


packages/dappmanager/src/calls/statsDiskGet.ts
```diff
@@ -0,0 +1,25 @@
...
+  const disk = await shellExec(`df / --block-size=1`);
+  return parseDiskStats(disk);
+}
+
+/**
+ * Parses the 'df /' bash output command
+ * @param disk string with disk usage info
+ */
+function parseDiskStats(disk: string): HostStatDisk {
+  const arr = disk.replace(/\n/g, " ").split(/\s+/);
```
dapplion, line 16: `parseDiskStats` should be unit tested like
```ts
it("Should parse df / output", () => {
  const dfOutput = `Filesystem                1K-blocks      Used Available Use% Mounted on
/dev/mapper/mint--vg-root 235782040 184988848  38746448  83% /`;
  const diskStats = parseDiskStats(dfOutput);
  expect(diskStats).to.deep.equal({
    filesystem: "/dev/mapper/mint--vg-root",
    bBlocks: "235782040",
    used: "184988848",
    available: "38746448",
    use: "83%",
    mountedOn: "/",
    useFraction: 0.342153612
  });
});
```

dapplion: I would also improve the parsing algorythm by splitting by rows first
```ts
const [headers, row] = dfOutput.split("\n");
const [filesystem, bBlocks, used, available, use, mountedOn] = row.split(/\s+/);
return {
  filesystem,
  bBlocks,
  used,
  available,
  use,
  mountedOn,
  useFraction: parseInt(used) / (parseInt(used) + parseInt(available))
};

```

packages/dappmanager/src/calls/statsMemoryGet.ts
```diff
@@ -0,0 +1,34 @@
...
+  const mem = await shellExec(`free / --bytes`);
+  return parseMemoryStats(mem);
+}
+
+/**
+ * Parses the 'free /' bash output command
+ * @param mem string with memory usage info
+ */
+function parseMemoryStats(memory: string): HostStatMemory {
+  const arr = memory.replace(/\n/g, " ").split(/\s+/);
```
dapplion, line 16: Same here, split rows first, then columns. Also unit test `parseMemoryStats` like `parseDiskStats`

packages/dappmanager/test/integration/fetchSystemData.test.int.ts
```diff
@@ -45,13 +45,22 @@ describe("Get system data", () => {
...
-  it("Should get host machine stats", async () => {
-    const result = await calls.getStats();
-    // Can't check actual returns since they vary
-    expect(result.cpu).to.include("%");
-    expect(result.memory).to.include("%");
-    expect(result.disk).to.include("%");
-  }).timeout(10 * 1000);
+  it("Should return parsed disk stats from host machine", done => {
+    calls
+      .statsDiskGet()
```
dapplion, line 50: Use await syntax like the previous test.
Doing
```
.then(done, done);
```
means that either if the promise rejects or resolves the test passes

packages/dappmanager/test/integration/fetchSystemData.test.int.ts
```diff
@@ -45,13 +45,22 @@ describe("Get system data", () => {
...
+      .then(done, done);
+  });
+  it("Should return parsed memory stats from host machine", done => {
+    calls
+      .statsMemoryGet()
+      .then(result => {
+        expect(result).to.be.ok;
+      })
+      .then(done, done);
+  });
```
dapplion, line 63: Add a third test for the CPU call

dapplion (changes requested):
Great job overall! Provided feedback and parsing functions and tests

packages/admin-ui/src/pages/dashboard/components/HostStats.tsx
```diff
@@ -2,41 +2,112 @@ import React, { useEffect } from "react";
...
   return (
     <div className="dashboard-cards">
-      {stats.data &&
-        Object.entries(stats.data).map(([id, percent]) => (
-          <StatsCard key={id} id={id} percent={percent} />
-        ))}
+      {cpuStats.data ? (
+        <StatsCard key={0} title={"cpu"} percent={cpuStats.data.used} />
+      ) : cpuStats.error ? (
+        <Loading steps={["Not possible to load cpu"]} />
```
dapplion, line 84: In case of error you want to show an error component instead of a loading indicator. You can use the `ErrorView` component for that

dapplion:
I've enabled JSON RPC over WebSockets on the mock server, so you can now run it locally and check the UI work.

packages/dappmanager/src/calls/statsDiskGet.ts
pablomendezroyo: Im having error with the deep equal test, I gess I should not use it. Since the commands `free `and `df `give different output depending on the host machine.

pablomendezroyo:
> I've enabled JSON RPC over WebSockets on the mock server, so you can now run it locally and check the UI work. The styling of the stats card component should match the chain card component: with the same colors and margin values

I already was able to run it locally and check the changes in the UI, just needed to setup two env variables in the env file (allow_all_ips and http_api_port) of the dappmanager.

Now 5 tests related with resolveDomain are failling in github, I hope/wish/guess it is because of the changes in the socket.

packages/dappmanager/test/utils/parseDiskStats.test.ts
```diff
@@ -0,0 +1,14 @@
...
+import { parseDiskStats } from "../../src/utils/parseDiskStats";
+
+describe("Util: parseDiskStats", async function() {
+  it("Should parse <df / --block-size=1> output", () => {
+    const dfOutput = `Filesystem         1B-blocks        Used    Available Use% Mounted on
+    /dev/nvme0n1p2 1006530654208 27418755072 927911559168   3% /`;
+
+    const diskStats = parseDiskStats(dfOutput);
+    expect(diskStats).to.not.be.empty;
+    expect(diskStats.filesystem).to.include("/"); // If filesystem argument has '/', the order should be correct.
```
dapplion: Here you must assert an exact result object will all the properties:
```ts
expect(diskStats).to.deep.equal({ ... data ... })
```
You control the input so you can assert the output without uncertainity.

packages/dappmanager/test/utils/parseMemoryStats.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe("Util: parseMemoryStats", async function() {
+  it("Should parse <free / --bytes> output", () => {
+    const freeOutput = `total        used        free      shared  buff/cache   available
+    Mem:    16535027712  5445357568  4503203840  1618006016  6586466304  9713278976
+    Swap:    2147479552           0  2147479552`;
+
+    const memoryStats = parseMemoryStats(freeOutput);
+    expect(memoryStats).to.not.be.empty;
+    expect(typeof memoryStats.useFraction).to.deep.equal("number"); // Only one number in memoryStats. If useFraction is number, the order is correct
```
dapplion: Same as above

dapplion (changes requested):

packages/dappmanager/test/utils/parseDiskStats.test.ts
pablomendezroyo: Yes, that is outdated. New PR are with the correct tests.

pablomendezroyo:
**I've added:**

- Function **okStatsCard():** Abstraction of the Ok element. this function returns Ok element depending on arguments. 
- **Style** for the **Ok** JSX element "InLine". I know that is not a good practice, but since **Ok** an element already predefined there are class names randomly generated (I think). I ve also seen that you give style to **Ok** element "inLine" in other parts of the code.

**Results**

Screenshot from 2020-11-12 11-50-19

Screenshot from 2020-11-12 12-02-29

Screenshot from 2020-11-12 11-52-56

Issue #420, opened 2020-10-28
Customize docker timeout


dapplion:
Some packages such as Geth require longer timeouts than the docker default of 10 seconds. For packages that need it, this timeout has to be longer

- [x] All docker and docker-compose actions that stop the container must have a customizable timeout option as a function parameter. These include docker stop, docker rm, docker-compose up, etc. Check https://github.com/dappnode/DNP_DAPPMANAGER/compare/v0.2.36...wait-to-kill as early work - https://github.com/dappnode/DNP_DAPPMANAGER/issues/420
- [x] Fetch a timeout parameter from a package manifest of a label attached to their container instance
- [x] Edit manifest schema with this timeout param https://github.com/dappnode/DAppNodeSDK/commit/4b2074510ac6281d7540ea676a955067dc0a75d1
- [x] Add dockerTimeout to Geth

PR #433, opened 2020-11-18
Dynamic dnp timeout from manifest

dapplion:
Developers can customize the timeout to stop a package containers with the manifest property
```ts
// "15min" | 3600
dockerTimeout?: string;
```

Fixes https://github.com/dappnode/DNP_DAPPMANAGER/issues/420


dapplion:
Ready for merge. Just needs testing with a package that uses the `dockerTimeout` prop

PR #472, opened 2020-12-03
Added package running status to docker compose up when updating

pablomendezroyo:
**Changes**

**1.** Added new packageInstalled type key "running"
**2.** Called listPackages() in packageInstall() to obtain info about packages
**3.** Get package status to avoid executing docker-compose-up against those stopped

### DIFF
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -417,6 +417,7 @@ export type InstalledPackageData = Pick<
   | "chain"
   | "domainAlias"
   | "canBeFullnode"
+  | "running"
 > & {
   containers: PackageContainer[];
 };
@@ -827,6 +828,7 @@ export interface InstallPackageData extends PackageRelease {
   // User settings to be applied after running
   fileUploads?: { [serviceName: string]: { [containerPath: string]: string } };
   dockerTimeout: number | undefined;
+  running: boolean | undefined;
 }
 
 // Must be in-sync with SDK types
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -827,8 +827,9 @@ export interface InstallPackageData extends PackageRelease {
   // User settings to be applied after running
   fileUploads?: { [serviceName: string]: { [containerPath: string]: string } };
   dockerTimeout: number | undefined;
+  // Package info
   runningContainers?: string[];
-  packageStopped?: boolean;
+  fullWorking?: boolean;
 }
 
 // Must be in-sync with SDK types
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -417,7 +417,6 @@ export type InstalledPackageData = Pick<
   | "chain"
   | "domainAlias"
   | "canBeFullnode"
-  | "running"
 > & {
   containers: PackageContainer[];
 };
@@ -828,7 +827,7 @@ export interface InstallPackageData extends PackageRelease {
   // User settings to be applied after running
   fileUploads?: { [serviceName: string]: { [containerPath: string]: string } };
   dockerTimeout: number | undefined;
-  running: boolean | undefined;
+  running: boolean;
 }
 
 // Must be in-sync with SDK types
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -827,7 +827,7 @@ export interface InstallPackageData extends PackageRelease {
   // User settings to be applied after running
   fileUploads?: { [serviceName: string]: { [containerPath: string]: string } };
   dockerTimeout: number | undefined;
-  running: boolean;
+  runningContainers?: string[];
 }
 
 // Must be in-sync with SDK types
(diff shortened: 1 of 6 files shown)


packages/dappmanager/src/modules/installer/getInstallerPackageData.ts
```diff
@@ -74,6 +90,7 @@ function getInstallerPackageData(
     compose: compose.output(),
     // User settings to be applied by the installer
     fileUploads: userSettings?.fileUploads,
-    dockerTimeout
+    dockerTimeout,
+    running
```
dapplion, line 94: We should know specifically which containers are running and which are not. If the package DMS has a prometheus and grafana services, where only prometheus is stopped; this code should
- Not start prometheus
- Start grafana

dapplion (changes requested):
Good direction! Now it needs to know which contains to start which to not

packages/dappmanager/src/modules/installer/runPackages.ts
```diff
@@ -54,13 +55,31 @@ export async function runPackages(
...
+      const removeOrphans = !pkg.isCore;
+
+      if (!pkg.running) {
+        await dockerComposeUp(pkg.composePath, {
+          noStart: pkg.running,
+          removeOrphans,
+          timeout: pkg.dockerTimeout
+        });
+      } else if (pkg.running) {
+        const packageInfo = await listPackage(pkg);
```
dapplion, line 69: Move this logic to packages/dappmanager/src/modules/installer/getInstallerPackageData.ts

packages/dappmanager/src/modules/installer/getInstallerPackageData.ts
```diff
@@ -10,18 +14,21 @@ export function getInstallerPackagesData({
...
   reqName: string;
+  packagesInfo: InstalledPackageData[];
 }): InstallPackageData[] {
   const packagesDataUnordered = releases.map(release =>
     getInstallerPackageData(
       release,
       userSettings[release.dnpName],
-      currentVersions[release.dnpName]
+      currentVersions[release.dnpName],
+      packagesInfo.find(pkg => pkg.dnpName === release.dnpName)
```
dapplion, line 31: If you only need one package there's no need to list all packages in `packages/dappmanager/src/calls/packageInstall.ts`

packages/admin-ui/src/common/types.ts
```diff
@@ -827,6 +827,7 @@ export interface InstallPackageData extends PackageRelease {
   // User settings to be applied after running
   fileUploads?: { [serviceName: string]: { [containerPath: string]: string } };
   dockerTimeout: number | undefined;
+  running: boolean;
```
dapplion, line 830: Change for
```ts
runningContainers: string[]
```

packages/dappmanager/src/modules/installer/runPackages.ts
```diff
@@ -59,27 +59,11 @@ export async function runPackages(
...
-        );
-        await dockerComposeUp(pkg.composePath, {
-          serviceNames: servicesRunning,
-          noStart: pkg.running,
-          removeOrphans,
-          timeout: pkg.dockerTimeout
-        });
-      }
+      await dockerComposeUp(pkg.composePath, {
+        serviceNames: pkg.runningContainers,
```
dapplion, line 63: We should really only use this option if not all services are running. If all services are running run docker-compose up without service arguments

packages/dappmanager/src/modules/installer/runPackages.ts
```diff
@@ -54,13 +55,20 @@ export async function runPackages(
...
-      const removeOrphans = !pkg.isCore;
 
       log(pkg.dnpName, "Starting package... ");
-      await dockerComposeUp(pkg.composePath, {
-        removeOrphans,
-        timeout: pkg.dockerTimeout
-      });
+      const removeOrphans = !pkg.isCore;
+
+      pkg.runningContainers
```
dapplion, line 62: Here https://github.com/dappnode/DNP_DAPPMANAGER/pull/472/files#diff-3e90a91864c6d48793b01d80c162240f3c31f12737695cba3b5039104553f96eR72
`runningContainers` will always be defined so this condition won't work

pablomendezroyo: Screenshot from 2020-12-07 10-36-58

runningContainers is possibly undefined

dapplion: According to typescript yes, but it will be undefined if the package is not found. If the package is found and all containers are running, `runningContainers` will be defined and contain all service names

dapplion:
To recap the functionality of the PR should be:
- Always create all the service's containers
- If some services are not running before the installation, start only the ones that are were running. Or none if all stopped
- If all services were running, use the command `docker-compose up` without any options

packages/dappmanager/src/calls/packageInstall.ts
```diff
@@ -74,12 +75,15 @@ export async function packageInstall({
         );
     }
 
+    const packageInfo = await listPackage({ dnpName: req.name });
```
dapplion, line 78: Use the noThrow variant or new packages won't be able to install

dapplion (changes requested):
Since types have changed, make sure to update test files accordingly

dapplion:
Superseded by https://github.com/dappnode/DNP_DAPPMANAGER/pull/530

PR #481, opened 2020-12-04
Enable multiple admin user logins

dapplion:
Allows to share a new set of username / password to external admins so the main admin of the DAppNode doesn't have to share its password to grant access. The created admin password is tied to a particular device and the credentials will be revoked once the device is removed


dapplion:
Looking good! @pablomendezroyo Would you mind testing this branch on your DAppNode? 
- Install new version
- Register, login, signout, change password with your first account
- Create a second admin and login with that password

pablomendezroyo:
> Looking good! @pablomendezroyo Would you mind testing this branch on your DAppNode?
> 
> * Install new version
> * Register, login, signout, change password with your first account
> * Create a second admin and login with that password

yep!

Issue #493, opened 2020-12-09
Strip ANSI from logs


dapplion:
Currently logs with ANSI characters look really bad. Either render it or strip it so at least it's possible to read it

eduadiez:
for instance:
```

[2m2020-12-09 09:16:40.180077[0m [[32m[1minfo     [0m] [1mMissing gas reserve           [0m [[34m[1mraiden.tasks[0m] [36mgreenlet_name[0m=[35mcheck_gas_reserve[0m [36mrequired_wei[0m=[35m150359413415199744[0m
WARNING
Your account's balance is below the estimated gas reserve of 0.150359413415199744 eth. This may lead to a loss of of funds because your account will be unable to perform on-chain transactions. Please add funds to your account as soon as possible.
[2m2020-12-09 09:16:48.880713[0m [[32m[1minfo     [0m] [1mSynchronizing blockchain events[0m [[34m[1mraiden.raiden_service[0m] [36mblocks_per_second[0m=[35m0.0662245316757279[0m [36melapsed[0m=[35m15.10014453400072[0m [36mgreenlet_name[0m=[35mAlarmTask._run node:0x4E88079eEBB4Bf3AB31F3C48A98B815B711B463F[0m [36mremaining_blocks_to_sync[0m=[35m0[0m [36mto_block[0m=[35m11417787[0m
[2m2020-12-09 09:17:14.972202[0m [[32m[1minfo     [0m] [1mSynchronizing blockchain events[0m [[34m[1mraiden.raiden_service[0m] [36mblocks_per_second[0m=[35m0.07665492398880855[0m [36melapsed[0m=[35m26.090952751997975[0m [36mgreenlet_name[0m=[35mAlarmTask._run node:0x4E88079eEBB4Bf3AB31F3C48A98B815B711B463F[0m [36mremaining_blocks_to_sync[0m=[35m0[0m [36mto_block[0m=[35m11417789[0m
[...]

PR #502, opened 2020-12-10
Strip ansi.

pablomendezroyo:
### **CHANGES**:
1. Added new libraries **has-ansi** and **strip-ansi**: both of them are popular and well supported.
2. Strip ansi in each line

Example 
Screenshot from 2020-12-10 11-45-56

Fixes https://github.com/dappnode/DNP_DAPPMANAGER/issues/493


dapplion (changes requested):
Can we use `strip-ansi` only? Otherwise we are running a RegExp twice per line

packages/dappmanager/src/modules/docker/utils.ts
```diff
@@ -33,7 +34,7 @@ import { PackageContainer } from "../../types";
 export function stripDockerApiLogsHeader(logs: string): string {
```
dapplion, line 34: Change function name `stripDockerApiLogsHeaderAndAnsi`

dapplion (changes requested):

dapplion (approved):

packages/dappmanager/src/modules/docker/dockerApi.ts
```diff
@@ -261,10 +261,10 @@ export async function logContainer(
   const res = await container.logs({ stdout: true, stderr: true, ...options });
   // Return is incorrectly typed as NodeJS.ReadableStream, but it's string
   /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  const data: string = (res as any) || "";
+  //const data: string = (res as any) || "";
```
dapplion, line 264: Why is this line commented? If res is undefined res.toString() will break

dapplion (changes requested):

dapplion (changes requested):
I've benchmark strip-ansi and doing it for each line instead of all at once it's 25! times slower. So strip-ansi must be called for all the text at once after being strip of docker's weird stuff
```
ansi at once: 8.098ms
ansi each line: 207.891ms
```

Source 

```ts
import stripAnsi from "strip-ansi";
import fs from "fs";

const data = fs.readFileSync(
  "./logs-dappnode-package-DAppNodePackage-ipfs-pinner.dnp.dappnode.eth.txt",
  "utf8"
);

console.time("ansi at once");
const noAnsi = stripAnsi(data);
console.timeEnd("ansi at once");

console.time("ansi each line");
for (const line of data.split("\n")) stripAnsi(line);
console.timeEnd("ansi each line");
```

dapplion (approved):

Issue #485, opened 2020-12-06
The status of packages with multiservice functionality is confusing


eduadiez:
If one of the packages of a multiservice is stopped, it seems like all are failing. This is maybe too alarmist since it may be okay for one package to be stopped on certain occasions.

Maybe we should show `Partial 2/4`in yellow or something similar, or if it's an `exit 0` shows it as stopped but ok.

<img width="1704" alt="Screenshot 2020-12-06 at 19 26 35" src="https://user-images.githubusercontent.com/20141918/101288839-fbbfcc00-37f8-11eb-938e-771852361dd0.png">
<img width="1685" alt="Screenshot 2020-12-06 at 19 01 18" src="https://user-images.githubusercontent.com/20141918/101288815-de8afd80-37f8-11eb-849b-7357b849fe39.png">
<img width="1686" alt="Screenshot 2020-12-06 at 19 01 25" src="https://user-images.githubusercontent.com/20141918/101288814-dcc13a00-37f8-11eb-9caf-213f071cd239.png">

eduadiez:
Taking the containers with an `exit 0` as stopped correctly seems very interesting. This allows having containers that can be stopped after the start depending on the configuration without warning the user.

dapplion:
With docker inspect one can get the precise status of a container and figure out if exited cleanly or not
```json
  "State": {
      "Status": "exited",
      "Running": false,
      "Paused": false,
      "Restarting": false,
      "OOMKilled": false,
      "Dead": false,
      "Pid": 0,
      "ExitCode": 0,
      "Error": "",
      "StartedAt": "2020-12-09T21:58:17.832941005Z",
      "FinishedAt": "2020-12-09T22:00:34.344620139Z"
  }
```
However, from the containers endpoint the status is less rich
```json
[
  {
    "State": "running",
    "Status": "Up 3 weeks"
  },
  {
    "State": "exited",
    "Status": "Exited (137) 19 hours ago"
  }
]
```
So the get the status with detail it should call inspect for each package or parse the Status text.

dapplion:
@eduadiez How should we name each of these package statuses?
- Clean exit, code == 0: 
- Not clean exit, code > 0: 
i.e. Stopped, Crashed

eduadiez:
First proposal:
| docker status | name | color |
| - | - | - |  
| created | created | white or light grey |
| restarting | restarting | yellow |
| running | running | green |
| paused | Paused | light green |
[...]

PR #521, opened 2020-12-18
Better multi-service package status

dapplion:
Instead of showing the worst state, 
- show all states in a compressed pill form with a legend only if necessary.
- Packages stop nicely will show as gray instead of red

Fixes https://github.com/dappnode/dnp_dappmanager/issues/485

## Main packages view

Screenshot from 2020-12-18 12-42-07

## Single package view (not expanded)

Screenshot from 2020-12-18 12-42-20

## Single package view (expanded)

Screenshot from 2020-12-18 12-42-32


ruvenni:
Big improvement imho. Go ahead! 👍🏽

ruvenni (approved):

PR #527, opened 2020-12-23
Fix logs UI bug with multi-services packages

pablomendezroyo:
**Main changes**

- New useState in _/admin-ui/src/pages/packages/components/Logs.tsx_: the new useState has been created to track whenever the serviceName selected has changed.

**Result**

Now the logs changed to display the new serviceName selected, even with the option "Auto refresh logs" disabled

Screenshot from 2020-12-23 09-12-42
Screenshot from 2020-12-23 09-12-27

### DIFF
--- a/packages/admin-ui/src/pages/packages/components/Logs.tsx
+++ b/packages/admin-ui/src/pages/packages/components/Logs.tsx
@@ -24,7 +24,6 @@ const validateLines = (lines: number) => !isNaN(lines) && lines > 0;
 export function Logs({ containers }: { containers: PackageContainer[] }) {
   const serviceNames = containers.map(c => c.serviceName).sort();
   const [serviceName, setServiceName] = useState(serviceNames[0]);
-  const [serviceNameHasChanged, setServiceNameHasChanged] = useState(false);
 
   // User options
   const [autoRefresh, setAutoRefresh] = useState(true);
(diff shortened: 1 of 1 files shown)


dapplion (changes requested):
Your solution is very valid but I think a similar result could be achieved without a new state variable. The error was not calling `logDnp` if autoRefresh is false, with something like this (maybe can be cleaned up further)
```ts
      if (autoRefresh) {
        const interval = setInterval(logDnp, refreshInterval);
        return () => {
          clearInterval(interval);
          unmounted = true;
        };
      } else {
        logDnp();
        return () => {
          unmounted = true;
        };
      }
```

pablomendezroyo:
> Your solution is very valid but I think a similar result could be achieved without a new state variable. The error was not calling `logDnp` if autoRefresh is false, with something like this (maybe can be cleaned up further)
> 
> ```ts
>       if (autoRefresh) {
>         const interval = setInterval(logDnp, refreshInterval);
>         return () => {
>           clearInterval(interval);
>           unmounted = true;
>         };
>       } else {
>         logDnp();
>         return () => {
>           unmounted = true;
>         };
>       }
> ```

Yeah much better yours!

packages/admin-ui/src/pages/packages/components/Logs.tsx
```diff
@@ -71,23 +69,19 @@ export function Logs({ containers }: { containers: PackageContainer[] }) {
         setAutoRefresh(false);
       }
     }
-    if (
-      (autoRefresh && validateLines(lines)) ||
-      serviceNameHasChanged === true
-    ) {
-      setLogs("fetching...");
+    if (autoRefresh) {
```
dapplion, line 72: Add `setLogs("fetching...");` again before the if, gives better UX

dapplion (changes requested):

dapplion (approved):

PR #590, opened 2021-01-22
Fix GLOBAL_ENVS_PATH

dapplion:


dapplion:
Tested with core and non-core packages and works well

PR #617, opened 2021-02-10
change hostname appearance UI

tropicar:
The main idea of this pull request is adding a new entry on the db, initialize this value, using this value on the UI instead of Hostname of the machine, create a component to modify this value(on system-profile).


packages/admin-ui/src/common/routes.ts
```diff
@@ -105,6 +105,10 @@ export interface Routes {
     toPath: string;
   }) => Promise<void>;
 
+  /** Set the dappnodeWebNameSet */
+
+  dappnodeWebNameSet: (kwargs: { dappnodeWebName: string }) => Promise<void>;
```
dapplion, line 110: You can simplify the types by doing
```ts
dappnodeWebNameSet: (dappnodeWebName: string) => Promise<void>;
```
I think there's no room for confusion here

packages/dappmanager/src/calls/systemInfoGet.ts
```diff
@@ -15,7 +15,7 @@ export async function systemInfoGet(): Promise<SystemInfo> {
     versionDataVpn: db.versionDataVpn.get(),
     // Network params
     ip: db.publicIp.get(),
-    name: db.serverName.get(),
+    dappnodeWebName: db.dappnodeWebName.get(),
```
dapplion, line 18: I would still return the `name` or `hostname`, and then add
```ts
dappnodeWebName: db.dappnodeWebName.get() || db.serverName.get()
```

dapplion (changes requested):

packages/admin-ui/src/pages/system/components/Advanced/ChangeDappnodeWebName.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  useEffect(() => {
+    setInput(dappnodeWebName);
+  }, [dappnodeWebName]);
+
+  function onChangeDappNodeWebName() {
+    withToastNoThrow(() => api.dappnodeWebNameSet(input), {
+      message: "Setting dappnode name ...",
+      onSuccess: "Regresh the page to see the changes"
+    });
```
dapplion, line 21: For the UI to be reactive it should re-fetch the systemInfo after this request. If you try the UI now, it does not update the name after changing it

dapplion (changes requested):

dapplion (approved):
Looks great!

PR #626, opened 2021-03-02
Update docker script

pablomendezroyo:
##### Options:
1. **Update docker engine** (Daemon, CLI and containerd) at the same time to a well known docker stable version.
2. **Update docker compose** to a well known stable version

##### Requirements to execute the script:
- Debian OS.
- Stretch, Bullyese or Buster
- Docker engine/compose current version LOWER than stable version: never allow a downgrade
- Architecture supported: amd64, arm64, armhf
- docker must exist

##### Critical:
- If dappmanager container is not running after docker upgrade a reboot will happened.

##### To do:
- [x] Finish frontend
- [x] Loof for a manner of getting docker and docker compose version from host 
- [x] List all supported docker versions by dappnode and defined them as metadata. Only allow upgrades from docker versions upgrades, add a warming for others.
- [x] Connect frontend and backend
- [x] Test script
- [ ] Add workflows to test script in multiple machines
- [x] Add check requirements in script and frontend
- [x] Add list react component
- [ ] Add unit tests 

View of requirements in frontend 

image


packages/dappmanager/hostScripts/docker_update.sh
```diff
@@ -0,0 +1,343 @@
...
+    echo "Debian version not supported by docker and/or dappnode. Versions supported: buster, stretch and bullseye"
+    exit 10
+  fi
+}
+
+function get_system_info(){
+  # info about /etc/os-release: https://www.freedesktop.org/software/systemd/man/os-release.html#:~:text=The%20%2Fetc%2Fos%2Drelease,like%20shell%2Dcompatible%20variable%20assignments.
+  if [ -f /etc/os-release ]; then
+    # freedesktop.org and systemd
+    . /etc/os-release
```
dapplion, line 112: Use `source`

packages/dappmanager/hostScripts/docker_update.sh
```diff
@@ -0,0 +1,343 @@
...
+      echo "docker compose version is stable"
+      exit 8
+    elif $(dpkg --compare-versions ${STABLE_DOCKER_COMPOSE_VERSION} "lt" ${DOCKER_COMPOSE_VERSION}); then 
+      echo "Illegal to downgrade docker compose"
+      exit 16
+    fi
+  fi
+
+  # check if architecture is supported by docker
+  if [ "$ARCHITECTURE" !=  "amd64" ] && [ "$ARCHITECTURE" !=  "arm64" ] && [ "$ARCHITECTURE" !=  "armhf" ]; then
```
dapplion, line 96: Consider using this https://stackoverflow.com/questions/21157435/bash-string-compare-to-multiple-correct-values

packages/dappmanager/hostScripts/docker_update.sh
```diff
@@ -0,0 +1,343 @@
+#!/bin/bash
```
dapplion, line 1: I strongly believe this script can be simplified a lot by splitting the docker and compose update logic in separate scripts. Then you will not need to parse arguments so much and handle branching logic

packages/dappmanager/hostScripts/docker_update.sh
```diff
@@ -0,0 +1,343 @@
...
+  # check if debian version is supported by docker
+  if [ "$VERSION_CODENAME" != "buster" ] && [ "$VERSION_CODENAME" != "stretch" ] && [ "$VERSION_CODENAME" != "bullseye" ]; then
+    echo "Debian version not supported by docker and/or dappnode. Versions supported: buster, stretch and bullseye"
+    exit 10
+  fi
+}
+
+function get_system_info(){
+  # info about /etc/os-release: https://www.freedesktop.org/software/systemd/man/os-release.html#:~:text=The%20%2Fetc%2Fos%2Drelease,like%20shell%2Dcompatible%20variable%20assignments.
+  if [ -f /etc/os-release ]; then
```
dapplion, line 110: You could just do at the top
```
source /etc/os-release || echo "/etc/os-release not found"
```
and won't need to re-assign variables

packages/dappmanager/hostScripts/docker_update.sh
```diff
@@ -0,0 +1,343 @@
...
+
+function post_install_check() {
+  # Docker-engine post installation check
+  if [ $INSTALL_OPTION == "engine" ]; then
+    DAPPMANAGER_CONTAINER="DAppNodeCore-dappmanager.dnp.dappnode.eth"
+
+    # If is container restarting wait 10 s
+    RESTARTING=$(docker inspect --format="{{.State.Restarting}}" $DAPPMANAGER_CONTAINER)
+    if [ $RESTARTING = true ]; then
+      sleep 10
```
dapplion, line 231: Why 10 seconds? Could it retry a few more times, like for 6 times check status and wait 10 seconds if not running

packages/dappmanager/hostScripts/docker_update.sh
```diff
@@ -0,0 +1,343 @@
...
+
+# Docker stable versions
+STABLE_DOCKER_ENGINE_VERSION="20.10.2" # Same for PKG and CLI
+STABLE_DOCKER_CONTAINERD_VERSION="1.4.3-1"
+
+# Docker compose stable versions
+STABLE_DOCKER_COMPOSE_VERSION="1.25.5"
+
+# Download
+WGET="wget -q -O"
```
dapplion, line 56: Not a fan of this alias, adds complexity in exchange to make the commands a tinny bit shorter

dapplion (changes requested):

packages/dappmanager/hostScripts/docker_update.sh
pablomendezroyo: My concern is that if the dappmanager is not running then the user will not have access to the UI, and could start doing crazy things if does not get response from the dappnode.

pablomendezroyo:
## Report before review

### Backend

**Main changes**
- 2 new host scripts: docker_engine_update.sh and `docker_compose_update.sh`
- Logic to determine requirements to update docker (compose and engine):
     - OS: allowed debian
     - OS version: allowed buster, bullseye and stretch
     - Architecture: allowed amd64, arm64 
     - Synchronization: docker server and CLI versions must be equal
     - Upgrade: docker version must be greater lower than docker engine stable version (defined in params)
     - Compatibility: docker compose must be greater or equal than docker compose stable version (defined in params)
- Parse scripts outputs logic
**Other changes**
- New types
- Routes to frontend

**Scripts**
##### `docker_engine_update.sh`
- Complex script. Critical installation. A lot of requirements to be checked
- options availabe: `--print-host-info` (returns update requirements in JSON) and `--install` (updates docker engine)
- Script update flow:
    1. **Check requirements** (while getting system info necessary to be checked): if requirements not fulfilled exit
    2. **Update docker engine** (according to system info)
    3. **Post install check**: check dappmanager container running and versions have been updated, otherwise reboot
##### `docker_compose_update.sh`
- simple script. Not a critical installation (just download binary to a folder)
- options availaible: `--version` (returns update requirements in JSON) and `--install` (updates docker compose)

**Notes**
- Updates requirements are double checked: in the dappmanager and in the scripts

### Frontend
**Main changes**
- Added card in **system > advanced**: update docker and docker compose
- Route with backend
- New custom component: List. 

**Notes**
- Updates buttons are only available if the requirements are fulfilled.

### Results
**UI**
- first-view
image
- Engine requirements
image
- Compose requirements (update only available if requirements fulfilled)
image



**Script test**

dapplion (approved):

dapplion (approved):

Issue #474, opened 2020-12-03
Review IPFS usage


dapplion:
In an effort to minimize the impact of disabling IPFS locally

- Review what endpoints are strictly necessary for package updates.
- Do not pin anything if the endpoint is external

PR #635, opened 2021-03-09
IPFS client refactor

dapplion:
- Move retry and cache logic out of the IPFS module
- Use object.get instead of object.stat to test a hash
- Add Ipfs class instead of single functions
- Simplify API

Fixes https://github.com/dappnode/DNP_DAPPMANAGER/issues/474


dapplion:
Tested, installs good but the DAppStore doesn't cache everything. Review

PR #644, opened 2021-03-12
Fix API port service endpoint

pablomendezroyo:


packages/dappmanager/src/params.ts
```diff
@@ -203,7 +203,7 @@ const params = {
   ETHFORWARD_PIN_ON_VISIT: true,
 
   // API endpoint check tcp ports. req: /publicIp?tcpPorts=8092,1194 | res: /[{tcpPort, status}]
-  PORT_SCANNER_SERVICE_URL: "http://173.249.41.199:3000",
+  PORT_SCANNER_SERVICE_URL: "http://213.136.76.177:3030",
```
dapplion, line 206: Try using the domain of the machine so you can use https

dapplion (approved):

PR #650, opened 2021-03-16
Collect host info from script on start

pablomendezroyo:


packages/admin-ui/src/common/types.ts
```diff
@@ -894,6 +894,12 @@ export interface PackageVersionData {
   commit?: string;
 }
 
+export interface TopicDataItem {
```
dapplion, line 897: Can you find a better name for this type? Topic Data and Item are very 3 generic works that are still generic together. The name should be scoped to the usage of this type

packages/dappmanager/src/calls/diagnose.ts
```diff
@@ -1,59 +1,24 @@
...
-  return shellExec(cmd)
-    .then((data: string) => ({ result: (data || "").trim() }))
-    .catch((e: Error) => ({ error: e.message }));
-}
-
-function uname(): {
-  result?: string;
-  error?: string;
-} {
+export async function diagnose(): Promise<TopicDataItem[]> {
```
dapplion, line 7: Why not just return `HostInfoScript`? What's the point of converting the object to an array?

packages/dappmanager/src/calls/diagnose.ts
```diff
@@ -1,59 +1,24 @@
...
   try {
-    return { result: [os.platform(), os.arch(), os.release()].join(", ") };
+    const hostInfo: HostInfoScript = await getHostInfoMemoized();
+    return parseHostInfo(hostInfo);
   } catch (e) {
-    return { error: e.message };
+    throw Error(`Error collecting host info on collect_host_info.sh: ${e}`);
```
dapplion, line 12: The user doesn't have to know the name of the script. We know. `Error collecting host info` is good enough

dapplion (changes requested):

dapplion (approved):

PR #656, opened 2021-03-19
Upgrade compose file version from 3.4 to 3.5

pablomendezroyo:


packages/admin-ui/src/common/types.ts
```diff
@@ -609,6 +609,7 @@ export interface ComposeService {
 export interface ComposeServiceNetwork {
   ipv4_address?: string;
   aliases?: string[];
+  name?: string;
```
dapplion, line 612: I don't think this type exists, can you review?

packages/dappmanager/src/modules/compose/unsafeCompose.ts
```diff
@@ -66,7 +66,7 @@ export function parseUnsafeCompose(
   const isCore = getIsCore(manifest);
 
   return cleanCompose({
-    version: composeUnsafe.version || "3.4",
+    version: composeUnsafe.version || "3.5",
```
dapplion, line 69: Here we must force the version to **at least** 3.5, otherwise it will break.

dapplion:
In this PR you should use the `name` property in the compose manipulation

packages/dappmanager/src/modules/compose/editor.ts
```diff
@@ -222,6 +222,13 @@ export class ComposeEditor {
     return parseUserSettings(this.compose);
   }
 
+  /**
+   * Returns compose network
+   */
+  getComposeNetwork(networkName: string): ComposeNetwork {
```
dapplion, line 228: You should return `: ComposeNetwork | null` and use null as indication that the network does not exist
```ts
return this.compose.networks?.[networkName] ?? null;
```

packages/dappmanager/src/modules/compose/unsafeCompose.ts
```diff
@@ -64,9 +65,11 @@ export function parseUnsafeCompose(
   const dnpName = manifest.name;
   const version = manifest.version;
   const isCore = getIsCore(manifest);
+  // Use of new compose feature "name"only available in version 3.5
+  // https://docs.docker.com/compose/compose-file/compose-file-v3/#name-1
 
   return cleanCompose({
-    version: composeUnsafe.version || "3.4",
+    version: sanitizeComposeFileVersion(composeUnsafe.version),
```
dapplion, line 72: What does it mean to sanitize a version? The function ensures that version is at least a minimum. Call it that

packages/dappmanager/src/modules/https-portal/utils.ts
```diff
@@ -21,3 +25,32 @@ export function removeNetworkAliasCompose(
   composeService.removeNetwork(networkName);
   compose.write();
 }
+
+/**
+ * Get compose file network and compose network settings from dncore_network
+ * And rewrites the compose with the core network edited
+ */
+export function coreNetworkMigration(container: PackageContainer): void {
```
dapplion, line 33: `migrateCoreNetworkInCompose`

packages/dappmanager/src/params.ts
```diff
@@ -109,6 +109,7 @@ const params = {
   DNP_PRIVATE_NETWORK_NAME: "dncore_network",
   DNP_PRIVATE_NETWORK_NAME_FROM_CORE: "network",
   DNP_EXTERNAL_NETWORK_NAME: "dnpublic_network",
+  MINIMUM_COMPOSE_VERSION: "3.5",
```
dapplion, line 112: Explain why 3.5 right above here

packages/dappmanager/src/modules/https-portal/utils.ts
```diff
@@ -49,7 +50,9 @@ export function coreNetworkMigration(container: PackageContainer): void {
   composeService.addNetwork(
     dncoreNetworkName,
     { ...serviceNetwork },
```
dapplion, line 52: Why spread within a new object? What's the purpose of it?

packages/dappmanager/src/modules/https-portal/utils.ts
```diff
@@ -49,7 +50,9 @@ export function coreNetworkMigration(container: PackageContainer): void {
...
-    { ...networkConfig, ...{ external: true, name: dncoreNetworkName } }
+    {
+      ...{ external: true, name: dncoreNetworkName }
```
dapplion, line 54: Same here

packages/dappmanager/test/integration/migration/migration.test.int.ts
```diff
@@ -0,0 +1,125 @@
+import { expect } from "chai";
+import fs from "fs";
+import { PackageContainer } from "../../../src/common";
+import { migrateCoreNetworkInCompose } from "../../../src/modules/https-portal/utils";
+import params from "../../../src/params";
+import shell from "../../../src/utils/shell";
+
+describe.only("Migration", () => {
```
dapplion, line 8: Remove .only

packages/dappmanager/src/modules/https-portal/migration.ts
```diff
@@ -18,18 +18,6 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
-  const dappmanagerAlias = getPrivateNetworkAlias({
-    dnpName: params.dappmanagerDnpName,
-    serviceName: params.dappmanagerDnpName
-  });
-  const dappmanagerEndpointConfig = await getEndpointConfig(
-    params.dappmanagerContainerName
-  );
-  if (hasAlias(dappmanagerEndpointConfig, dappmanagerAlias)) {
-    return logs.info(`Alias migration already done`);
-  }
```
dapplion, line 31: Why remove this? We would run `docker inspect` on all running containers every time the dappmanager starts

packages/dappmanager/test/integration/migrationCompose/migration.test.int.ts
```diff
@@ -0,0 +1,161 @@
...
+import { getPrivateNetworkAlias } from "../../../src/domains";
+import {
+  addNetworkAliasCompose,
+  migrateCoreNetworkInCompose
+} from "../../../src/modules/https-portal/utils";
+import params from "../../../src/params";
+import shell from "../../../src/utils/shell";
+
+describe("Migration", () => {
+  const container: PackageContainer = {
```
dapplion, line 13: Use `mockContainer` to only specify the info you really need to customize, i.e.
```ts
const container: PackageContainer = {
  ...mockContainer,
  dnpName: "custom"
}

```

packages/dappmanager/test/integration/migrationCompose/migration.test.int.ts
```diff
@@ -0,0 +1,161 @@
...
+      `docker network disconnect ${dncoreNetwork} ${containerName} --force`
+    );
+    // Remove network
+    await shell(`docker network rm dncore_network`);
+    // Remove container
+    await shell(`docker rm ${containerName} --force`);
+    // Remove image
+    await shell(`docker image rm ${randomImage}`);
+    // Remove dir
+    await shell(`rm -rf ${testMigrationPath}/test-migration`);
```
dapplion, line 155: Use shellSafe on all of this. If the test fails early those may not be created

packages/dappmanager/test/integration/resilienceFeatures.test.int.ts
```diff
@@ -27,20 +27,24 @@ describe("Resilience features, when things go wrong", function() {
...
   before("Create DAppNode docker network", async () => {
-    const dncoreNetwork = params.DNP_PRIVATE_NETWORK_NAME;
     const networkExists = await shell(
       `docker network ls --filter name=${dncoreNetwork} -q`
     );
     if (!networkExists) await shell(`docker network create ${dncoreNetwork}`);
   });
 
+  after("Remove dncore_network", async () => {
+    await shell(`docker network rm ${dncoreNetwork}`);
```
dapplion, line 45: shellSafe

packages/dappmanager/src/modules/https-portal/migration.ts
pablomendezroyo: because there would be an unordered update of docker composes of core packages. I guess the roadmap is to finally have all composes updated with the version 3.5, the alias, and the new network configuration.

In this PR I take the oportunity to also update the dappmanager compose so the migration cannot deppend on the dappmanager. It cannot deppend on any core package since its compose will be randomly updated and we cannot be sure

packages/dappmanager/src/modules/https-portal/migration.ts
```diff
@@ -34,6 +34,19 @@ export async function addAliasToRunningContainersMigration(): Promise<void> {
...
+      // Wifi and VPN containers needs a refresh connect due to its own network configuration
+      if (
+        container.containerName === params.vpnContainerName ||
+        container.containerName === params.wifiContainerName
+      ) {
+        await shell(`docker rm ${containerName} --force`);
+        await shell(
+          `docker-compose -f ${params.DNCORE_DIR}/${
+            container.dnpName.split(".")[0]
+          } up -d`
```
dapplion, line 46: You should re-use the docker-compose wrappers from modules/docker https://github.com/dappnode/DNP_DAPPMANAGER/blob/cb58825c4a/packages/dappmanager/src/modules/docker/compose/cli.ts

packages/dappmanager/src/modules/https-portal/migration.ts
dapplion: Then what is expected to happen when you run `addAliasToRunningContainersMigration` after a successful migration?

packages/dappmanager/src/modules/https-portal/migration.ts
```diff
@@ -18,33 +20,34 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
+        container.containerName === params.wifiContainerName
+      ) {
+        await shell(`docker rm ${containerName} --force`);
+        await dockerComposeUp(
+          `${params.DNCORE_DIR}/docker-compose-${
+            container.dnpName.split(".")[0]
+          }.yml`
+        );
+        continue;
+      }
```
dapplion, line 50: if (vpn) { do vpn stuff } else { do other stuff }

packages/dappmanager/src/modules/https-portal/utils.ts
```diff
@@ -26,6 +26,16 @@ export function removeNetworkAliasCompose(
   compose.write();
 }
 
+export function composeVersionMigration(container: PackageContainer): void {
```
dapplion, line 29: This function on it's own is confusing, who-ever uses the `name` feature should ensure that the docker-compose version is at least 3.5

packages/dappmanager/src/modules/https-portal/migration.ts
pablomendezroyo: discussed

dapplion (approved):

PR #673, opened 2021-03-29
Fix bug in migration: check if aliases is defined in compose

pablomendezroyo:
Error in migration process:
```
ERROR [modules/https-portal/migration:63] Error adding alias to container DAppNodeCore-dappmanager.dnp.dappnode.eth TypeError: Cannot read property 'aliases' of undefined
    at migrateCoreNetworkAndAliasInCompose (/usr/src/app/webpack:/src/modules/https-portal/migration.ts:121:20)
```

### DIFF
--- a/packages/dappmanager/src/modules/https-portal/migration.ts
+++ b/packages/dappmanager/src/modules/https-portal/migration.ts
@@ -116,10 +116,10 @@ export function migrateCoreNetworkAndAliasInCompose(
     params.DNP_PRIVATE_NETWORK_NAME
   );
   if (
-    composeNetwork?.name === params.DNP_PRIVATE_NETWORK_NAME &&
-    composeNetwork?.external &&
-    serviceNetwork.aliases &&
-    serviceNetwork.aliases.includes(alias)
+    !serviceNetwork || // serviceNetwork will not exist in composes already migrated
+    (composeNetwork?.name === params.DNP_PRIVATE_NETWORK_NAME &&
+      composeNetwork?.external &&
+      serviceNetwork.aliases?.includes(alias))
   )
     return;
 
--- a/packages/dappmanager/src/modules/https-portal/migration.ts
+++ b/packages/dappmanager/src/modules/https-portal/migration.ts
@@ -16,7 +16,7 @@ import { PackageContainer } from "../../types";
 import { parseServiceNetworks } from "../compose/networks";
 import { ComposeNetwork, ComposeServiceNetwork } from "../../common";
 import semver from "semver";
-import { sanitizeVersion } from "../../utils/sanitizeVersion";
+import { parseComposeSemver } from "../../utils/sanitizeVersion";
 
 /** Alias for code succinctness */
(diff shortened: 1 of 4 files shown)


packages/dappmanager/src/modules/https-portal/migration.ts
```diff
@@ -118,7 +118,8 @@ export function migrateCoreNetworkAndAliasInCompose(
   if (
     composeNetwork?.name === params.DNP_PRIVATE_NETWORK_NAME &&
     composeNetwork?.external &&
-    serviceNetwork.aliases?.includes(alias)
+    serviceNetwork.aliases &&
+    serviceNetwork.aliases.includes(alias)
```
dapplion, line 122: This doesn't fix the bug, the new syntax is equivalent to the old one. The issue is that apparently `serviceNetwork` is not properly typed and can be nullish

packages/dappmanager/src/modules/https-portal/migration.ts
```diff
@@ -138,3 +144,36 @@ export function migrateCoreNetworkAndAliasInCompose(
...
+  });
+  const minimumComposeVersionCleaned = semver.clean(
+    params.MINIMUM_COMPOSE_VERSION + ".0",
+    {
+      loose: true
+    }
+  );
+
+  if (!composeVersionCleaned || !minimumComposeVersionCleaned)
+    throw Error("Docker compose file version cannot be used by semver");
```
dapplion, line 166: Is it really necessary to throw an error here? Would that prevent the migration of other packages? If you really want to throw, consider adding a try / catch in the for loop iterating over packages

packages/dappmanager/src/modules/https-portal/migration.ts
```diff
@@ -138,3 +144,36 @@ export function migrateCoreNetworkAndAliasInCompose(
...
+  // semver can only work with version of 3 numbers. e.g: 3.5.0
+  const composeVersionCleaned = semver.clean(composeVersion + ".0", {
+    loose: true
+  });
+  const minimumComposeVersionCleaned = semver.clean(
+    params.MINIMUM_COMPOSE_VERSION + ".0",
+    {
+      loose: true
+    }
+  );
```
dapplion, line 163: Consider adding a function parseComposeVersion() to de-duplicate the logic and explain why is this necessary

packages/dappmanager/src/modules/https-portal/migration.ts
pablomendezroyo: semver.clean may throw a null version. I guess it is almost impossible taking into account the docker compose file version format (2 digits and I add a 0 at the end), but just in case

dapplion: Then isolate that into `parseComposeVersion()` with signature
```ts
function parseComposeVersion(v: string): string
```
and it throws if clean return falsy

dapplion (approved):

PR #681, opened 2021-04-01
Bump to Node 12 LTS

dapplion:
Version 10 will stop being maintained this month https://nodejs.org/en/about/releases/


Jakic007:
Looks good, but it would be better not to use `ubuntu-latest` and put 20.04 explicitly.

PR #690, opened 2021-04-02
Fix auto updates return in loop bug

dapplion:
Fixes a regression introduced in DAPPMANAGER version `v0.2.38` which prevents auto-updating non-system packages if some package previous to it is already updated.

https://github.com/dappnode/DNP_DAPPMANAGER/blob/845a220da45754d2455a18692178e23d6c55258f/packages/dappmanager/src/daemons/autoUpdates/updateMyPackages.ts#L42

Introduced in commit: https://github.com/dappnode/DNP_DAPPMANAGER/pull/563/commits/845a220da45754d2455a18692178e23d6c55258f#diff-ad2a1d84e4dc97132ac60340f7678d281cf6fa16cb18f6f2f61ce5ab503a073fR42

Depending on the ordering of the containers returned by the docker API this bug may prevent all non-core packages from being updated or only some.

---

I've reviewed all return statements in the repo and there's no other instance of a similar bug

### DIFF
--- a/packages/dappmanager/src/daemons/autoUpdates/updateMyPackages.ts
+++ b/packages/dappmanager/src/daemons/autoUpdates/updateMyPackages.ts
@@ -25,21 +25,27 @@ export async function checkNewPackagesVersion(
   const dnps = await listPackages();
 
   for (const { dnpName, isDnp, version: currentVersion } of dnps) {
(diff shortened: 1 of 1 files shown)


dapplion:
Confirm that it fixes the bug and auto-updates work fine for all packages

PR #694, opened 2021-04-05
Update docker engine through host service

pablomendezroyo:
### Main changes
- New volume to track services added: `"/etc/systemd/system/:/usr/src/app/DNCORE/services/host/"`
- New service added to update docker engine
- Copy services on dappmanager startup

### Concerns
- Volume created might be critical due to the importance to the `/etc/systemd/system` host folder, it might be also necessary due to the easy handle of new services added in dappnode and/or possible changes to the existing ones.
- After docker engine update containers will restart and **connection will be lost** for a little while.

### UPDATE:
- Instead of using volume (it is potentiallly dangerous for the host), just copy the service when attemping to run it


packages/dappmanager/src/modules/hostServices/runServices.ts
```diff
@@ -0,0 +1,28 @@
...
+  args = ""
+): Promise<string> {
+  const servicePath = path.resolve(
+    params.HOST_SERVICES_SOURCE_DIR,
+    serviceName
+  );
+  if (!fs.existsSync(servicePath))
+    throw Error(`Host service ${serviceName} not found`);
+
+  return await shellHost(`systemctl start ${serviceName} ${args}`);
```
3alpha, line 27: Keep in mind that if *.service file is ever edited, before this command you need to run `systemctl daemon-reload`. I don't know whether that will be ever the case, but keep it in mind.

pablomendezroyo: god point, I will edit the PR taking this into account

pablomendezroyo: For this specific service, `systemctl enable update-docker-engine.service` is not executed since the behaviour desired is to be executed only on demmand and not on boot, so `systemctl daemon-reload` would not be necessary

Anyway might be useful for any other service there might be added in the future so I will add it

docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       - "/run/dbus/system_bus_socket:/run/dbus/system_bus_socket"
       - "dappmanagerdnpdappnodeeth_data:/usr/src/app/dnp_repo/"
       - "/usr/src/dappnode/DNCORE/:/usr/src/app/DNCORE/"
+      - "/etc/systemd/system/:/usr/src/app/DNCORE/services/host/"
```
dapplion, line 18: @3alpha Do you see any security issues with mapping this path?

pablomendezroyo: The most dangerous thing I have found is to delete the docker volume, that would remove the host services which is potentially dangerous

dapplion (approved):

PR #701, opened 2021-04-06
Replace https error view with banner

pablomendezroyo:


packages/admin-ui/src/pages/packages/components/Network/HttpsMappings.tsx
```diff
@@ -107,6 +107,19 @@ export function HttpsMappings({
...
+    return (
+      <>
+        <p>
+          To do HTTPS networking is necessary to install the system package
+          HTTPS.
+        </p>
+        <br />
+        <a href="http://my.dappnode/#/installer/https-portal.dnp.dappnode.eth">
+          Get HTTPS package
+        </a>
```
dapplion, line 120: The ADMIN UI is React single page application and all navigation is done programatically. Look around to found examples with `react-router` components

packages/admin-ui/src/pages/packages/components/Network/HttpsMappings.tsx
```diff
@@ -107,6 +107,19 @@ export function HttpsMappings({
     }
   }
 
+  if (mappings.error && mappings.error.message.includes("ENOTFOUND"))
```
dapplion, line 110: This error detection is too broad. The DAPPMANAGER should be able to detect that the HTTPs portal is not installed and communicate that to the Admin UI in a clear way

dapplion (approved):

PR #708, opened 2021-04-09
add predefined exposable services

tropicar:
Fix the exposable predefined service goerli and adding kovan services.


tropicar:
I had problem doing git rebase on this branch,

PR #742, opened 2021-04-16
Wifi refactor

pablomendezroyo:


packages/admin-ui/src/__mock-backend__/data/dnps/multiService.ts
```diff
@@ -1,4 +1,4 @@
-import { PortProtocol } from "types";
+import { PortProtocol } from "./types";
 import { MockDnp } from "./types";
```
dapplion, line 2: Merge all the double imports into a single statement

packages/admin-ui/src/__mock-backend__/data/dnps/types.ts
```diff
@@ -7,6 +7,8 @@ import {
   PackageContainer
 } from "../../../common";
 
+export { PortProtocol } from "../../../common";
```
dapplion, line 10: Merge duplicated imports in a single statement

packages/admin-ui/src/common/routes.ts
```diff
@@ -528,6 +530,11 @@ export interface Routes {
    */
   ipPublicGet: () => Promise<PublicIpResponse>;
 
+  /**Get wifi credentials */
```
dapplion, line 533: `/** Get wifi credentials */`

packages/admin-ui/src/pages/wifi/components/WifiCredentials.tsx
```diff
@@ -0,0 +1,121 @@
...
+      );
+  }
+
+  return (
+    <>
+      {wifiCredentials.error ? (
+        <ErrorView error={wifiCredentials.error} />
+      ) : wifiCredentials.isValidating ? (
+        <Loading steps={["Loading wifi credentials"]} />
+      ) : wifiCredentials.data ? (
```
dapplion, line 69: Always render:
1. data
2. elif error
3. elif loading

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
```diff
@@ -0,0 +1,159 @@
...
+    </>
+  );
+}
+
+function ActionsWifi({ wifiStatus }: { wifiStatus: ContainerState }) {
+  async function restartWifi() {
+    await new Promise<void>(resolve => {
+      confirm({
+        title: `Restart wifi service`,
+        text: `Warming, you may loose connection to your dappnode during wifi restart`,
```
dapplion, line 68: `you connection may break if you connected to your DAppNode's WIFI`

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
```diff
@@ -0,0 +1,159 @@
...
+      )
+    );
+  }
+
+  async function pauseWifi() {
+    if (wifiStatus === "running")
+      await new Promise<void>(resolve => {
+        confirm({
+          title: `Pause wifi service`,
+          text: `Warming, you may loose connection to your dappnode if stopping this service`,
```
dapplion, line 87: `Warning, if you are connected to your DAppNode's WIFI you may lose access to your DAppNode. Make sure to have an alternative way to connect to it, like a VPN connection.

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
```diff
@@ -0,0 +1,159 @@
...
+        wifiDnpName
+      )
+    );
+  }
+
+  return (
+    <div className="wifi-actions">
+      <Button
+        variant="warning"
+        onClick={() => restartWifi().catch(console.error)}
```
dapplion, line 105: Don't do catch here, just do
```ts
onClick={restartWifi}
```
if you need to handle errors do it in the `restartWifi` fn

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
```diff
@@ -0,0 +1,159 @@
...
+    };
+  }, [wifiReport]);
+
+  return (
+    <>
+      {wifiReport.error ? (
+        <ErrorView error={wifiReport.error} />
+      ) : wifiReport.isValidating ? (
+        <Loading steps={["Loading wifi service"]} />
+      ) : wifiReport.data ? (
```
dapplion, line 137: data, error, loading

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return {
+    containerState: wifiDnp.state,
+    info,
+    report
+  };
+}
+
+const composeWifi = new ComposeFileEditor(params.wifiDnpName, true);
```
dapplion, line 60: This will be executed once per DAPPMANAGER run on start. Is this meant to be here?

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,76 @@
...
+    case "created":
+      info = "Wifi has not been initialized yet";
+    case "running":
+      info = "Wifi is currently running";
+    case "restarting":
+      info = "Wifi restarting. Wait until it starts";
+    case "dead":
+      info =
+        "Wifi service dead, you must manually remove it and install it again";
+      report = { lastLog: await getWifiLastLog(), exitCode: wifiDnp.exitCode };
```
dapplion, line 44: What will the UI do with this log? You should parse it here and try to understand it on the backend

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,76 @@
...
+const WIFI_KEY_SSID = "SSID";
+const WIFI_KEY_PASSWORD = "WPA_PASSPHRASE";
+
+/**
+ * Get last log from wifi container
+ * Contains useful information for the user, check echos before exit codes:
+ * https://github.com/dappnode/DNP_WIFI/blob/master/build/wlanstart.sh
+ */
+async function getWifiLastLog(): Promise<string> {
+  return await logContainer(params.wifiContainerName, { tail: 1 });
```
dapplion, line 16: You can improve this function by checking the time the container stopped and requesting the logs from that time and now. Then look for the last log that matches some expected pattern.

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const composeWifi = new ComposeFileEditor(params.wifiDnpName, true);
+
+/**
+ * Returns wifi credentials
+ */
+export async function wifiCredentialsGet(): Promise<CurrentWifiCredentials> {
+  const wifiService = composeWifi.services()[params.wifiDnpName];
+  const wifiEnvs = wifiService.getEnvs();
+  if (!wifiEnvs[WIFI_KEY_SSID] || !wifiEnvs[WIFI_KEY_PASSWORD])
```
dapplion, line 68: What if they are empty strings?

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
```diff
@@ -0,0 +1,159 @@
...
+        () => api.packageStartStop({ dnpName: wifiDnpName }),
+        wifiDnpName
+      )
+    );
+  }
+
+  return (
+    <div className="wifi-actions">
+      <Button
+        variant="warning"
```
dapplion, line 104: We have no button in the UI with warning variant, so it breaks with the overall style. You should use outline-secondary or dappnode. However the UI could be simplifed I think. Users may just want to enable or disable the WIFI. So just a big switch will be enough.

If something goes wrong users can still go to packages > system > WIFI and debug the situation

packages/admin-ui/src/__mock-backend__/data/dnps/types.ts
pablomendezroyo: one is an import, the other one an export

packages/dappmanager/src/calls/wifi.ts
pablomendezroyo: I do not see that as an option since there is added default credentials for the wifi. Also the user have a constraint in the UI that does not allow to set empty credentials

packages/dappmanager/src/calls/wifi.ts
pablomendezroyo: AFAIK in most of cases when a container stops, the last log is the error log explaining why the container stopped. In this specific use case, the wifi container exits at least three times with a meaningful log
https://github.com/dappnode/DNP_WIFI/blob/00685cf6160df4b8daeabd77246cb0af04a513f8/build/wlanstart.sh#L5
https://github.com/dappnode/DNP_WIFI/blob/00685cf6160df4b8daeabd77246cb0af04a513f8/build/wlanstart.sh#L34
https://github.com/dappnode/DNP_WIFI/blob/00685cf6160df4b8daeabd77246cb0af04a513f8/build/wlanstart.sh#L52

Also, I do not see the point of requesting the time of when the container stopped. After that time there will not be logs since the container is not running anymore

I could add parsing functions to interpret logs

packages/dappmanager/src/calls/wifi.ts
pablomendezroyo: I have added a markdown render, only stderr will be returned

packages/admin-ui/src/__mock-backend__/wifi.ts
```diff
@@ -9,7 +9,10 @@ export const wifi: Pick<Routes, "wifiCredentialsGet" | "wifiReportGet"> = {
       containerState: "exited",
       info: "Wifi service exited due to an internal error",
       report: {
-        lastLog: `[Warning] No interface found after 60s, stopping gracefully`,
+        lastLog: "[Warning] No interface found after 60s, stopping gracefully".replace(
+          /\[.*?\]/g,
+          ""
+        ),
```
dapplion, line 15: What is this regex doing? Since this is a static text could you just use the result?

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
```diff
@@ -48,16 +50,18 @@ function WifiAlert({ wifiReport }: { wifiReport: WifiReport }) {
...
-        {wifiReport.info}.{" "}
-        <RenderMarkdown
-          source={`~~~js\n${wifiReport.report?.lastLog}\n~~~` || ""}
-        />
-      </p>
-      <hr />
+      <Alert.Heading>{wifiReport.info}</Alert.Heading>
+      {wifiReport.report ? (
+        <p>
+          <RenderMarkdown
```
dapplion, line 60: Why do we want to render markdown here?

packages/admin-ui/src/__mock-backend__/wifi.ts
dapplion: Oh I see. Hmm, is this user-friendly enough? Just removing this tag and showing the message on screen is good? What is the user expected to do when he/she sees
```
No interface found after 60s, stopping gracefully
```
?

packages/admin-ui/src/pages/wifi/components/WifiService.tsx
pablomendezroyo: Last error log from the wifi container when `exited` or `dead`

I tried to be consistent since I cannot prevent every single log from wifi package. Also if hardcoding known exit logs it may be outdated in upcoming releases

packages/admin-ui/src/__mock-backend__/wifi.ts
pablomendezroyo: I tried to make it as friendly as possible and be as consistent as possible. To do that I return last error log when wifi container was exited/dead and render it as markdown code

I cannot prevent how is going to be every single error log from the wifi container, also hardcode them may be risky due to it can get unupdated soon.

Im open to any other suggestion

packages/admin-ui/src/__mock-backend__/data/dnps/lightningNetwork.ts
```diff
@@ -1,4 +1,4 @@
-import { PortProtocol } from "types";
+import { PortProtocol } from "./types";
 import { bitcoin } from "./bitcoin";
 import { MockDnp } from "./types";
```
dapplion, line 3: Merge `./types` imports

dapplion: Not only here but review all new code for duplicated imports

packages/admin-ui/src/common/types.ts
```diff
@@ -32,6 +32,22 @@ export interface ExposableServiceMapping extends ExposableServiceInfo {
   exposed: boolean;
 }
 
+// Wifi
+
+export interface WifiReport {
+  wifiContainer: PackageContainer;
```
dapplion, line 38: Not a fan of sending the wifiContainer data here. Some parts of the UI use `packagesGet` and this uses something else. It can lead to inconsistencies this view says the package is stopped and then packages > WIFI says that the package is running. Just fetch `packagesGet` in the view along with the rest of the `WifiReport` data structure.

packages/admin-ui/src/pages/wifi/components/WifiStatus.tsx
```diff
@@ -0,0 +1,150 @@
...
+// Types
+import { ContainerState, WifiReport } from "types";
+// css
+import "./wifi.scss";
+import { StateBadgeContainer } from "pages/packages/components/StateBadge";
+import { MdWifi } from "react-icons/md";
+// Utils
+import { prettyDnpName } from "utils/format";
+
+const wifiDnpName = "wifi.dnp.dappnode.eth";
```
dapplion, line 25: Don't add any important constants in deep files. All constants like this should go into the UI's params file

packages/admin-ui/src/pages/wifi/components/WifiStatus.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+function WifiInfo({ wifiStatus }: { wifiStatus: ContainerState }) {
+  const dappnodeIdentity = useSelector(getDappnodeIdentityClean);
+  return (
+    <p>
+      Connect to the Wi-Fi hostpot exposed by your DAppNode using your{" "}
+      <a href="http://my.dappnode/#/wifi/credentials">credentials</a>. More
```
dapplion, line 32: Never write links directly. All links should be variables and declared in the links section of params https://github.com/dappnode/DNP_DAPPMANAGER/blob/6838be01fbd24e346c8c6e8c6a2a0940afcd939e/packages/admin-ui/src/params.ts#L65

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,92 @@
+import { ComposeFileEditor } from "../modules/compose/editor";
+import { logContainer } from "../modules/docker/api";
+import { listContainer } from "../modules/docker/list";
+import { CurrentWifiCredentials, WifiReport } from "../types";
+import params from "../params";
+
+const WIFI_KEY_SSID = "SSID";
+const WIFI_KEY_PASSWORD = "WPA_PASSPHRASE";
```
dapplion, line 8: All constants must go to the params file

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Return wifi report
+ */
+export async function wifiReportGet(): Promise<WifiReport> {
+  const wifiContainer = await listContainer({
+    containerName: params.wifiContainerName
+  });
+  if (!wifiContainer)
+    throw Error(
+      `Wifi is not present. Container name: ${params.wifiContainerName}`
```
dapplion, line 19: What does it mean "Wifi is not present"? If a container is not in the list means that there is no container with that name. In DAppNode terminology it means that it's not installed. Think if you were a user and find yourself in that situation, what are the next steps?

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,92 @@
...
+  switch (wifiContainer.state) {
+    case "created":
+      info = "Wifi has not been initialized yet";
+    case "running":
+      info = "Wifi is currently running";
+    case "restarting":
+      info = "Wifi restarting. Wait until it starts";
+    case "dead":
+      info =
+        "Wifi service dead, you must manually remove it and install it again";
```
dapplion, line 37: Is it even possible for users to do that? Can they remove the WIFI container and install it again? How would they do that

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Returns wifi credentials
+ */
+export async function wifiCredentialsGet(): Promise<CurrentWifiCredentials> {
+  const composeWifi = new ComposeFileEditor(params.wifiDnpName, true);
+  const wifiService = composeWifi.services()[params.wifiDnpName];
+  const wifiEnvs = wifiService.getEnvs();
+  if (
+    typeof wifiEnvs[WIFI_KEY_SSID] === "undefined" ||
```
dapplion, line 67: You can do `wifiEnvs[WIFI_KEY_SSID] === undefined` which is much more typesafe. `typeof` should never be used unless in very specific scenarios

packages/dappmanager/src/calls/wifi.ts
```diff
@@ -0,0 +1,92 @@
...
+
+// Utils
+
+/**
+ * Get last log from wifi container
+ * Contains useful information for the user, check echos before exit codes:
+ * https://github.com/dappnode/DNP_WIFI/blob/master/build/wlanstart.sh
+ */
+async function getWifiLastLog(): Promise<string> {
+  return await logContainer(params.wifiContainerName, { stderr: true });
```
dapplion, line 87: Does it really get last log, or all the logs?

packages/admin-ui/src/pages/wifi/components/wifi.scss
```diff
@@ -0,0 +1,47 @@
...
+  border-radius: 5px;
+  background-color: #f5f5f552;
+  /* Text styling */
+  display: block;
+  white-space: pre;
+  font-size: 75%;
+  font-family: "Inconsolata", monospace;
+  overflow: auto;
+  max-height: 500px;
+}
```
dapplion, line 47: Please review all these styles and ensure they are strictly necessary for your new views

packages/dappmanager/src/calls/wifi.ts
pablomendezroyo: dead status would be a really weird and critical situation that must never happen, I have only seen that situation when trying to remove containers and these containers were partially removed.

Anyway if that happened for any reason, would be useful info for the user to report and get back the wifi

packages/dappmanager/src/calls/wifi.ts
pablomendezroyo: fixed with `tail: 1`

dapplion (approved):

PR #748, opened 2021-04-27
security updates

pablomendezroyo:


packages/admin-ui/src/common/routes.ts
```diff
@@ -512,6 +512,11 @@ export interface Routes {
    */
   systemInfoGet: () => Promise<SystemInfo>;
 
+  /**
+   * Executes security updates on host
+   */
+  updateSecurity: () => Promise<string>;
```
dapplion, line 518: Better name would be `runHostSecurityUpdates`

packages/dappmanager/hostScripts/security_update.sh
```diff
@@ -0,0 +1,147 @@
...
+
+
+##################
+#2.UPDATE-UPGRADE#
+##################
+
+function update () {
+  echo -e "\e[32mUpdating\e[0m" >> $LOG_FILE 
+  # 1>/dev/null 2>>$LOG_FILE: will redirect stdout to null and stderr to log file
+  apt-get update 1>/dev/null 2>>$LOG_FILE || { echo "error on apt-get update" 2>&1 | tee -a $LOG_FILE ; exit 1; }
```
3alpha, line 83: Why not log both stdin and stderr?

pablomendezroyo: AFAIK the important info would be just the error logs. When testing this script there were tons of logs saying a lot of stuff that I did not consider much important

packages/admin-ui/src/common/routes.ts
dapplion: bump

pablomendezroyo: fixed on last commit

packages/dappmanager/test/integration/hostScripts.test.int.ts
```diff
@@ -28,8 +28,7 @@ describe.only("Host scripts", () => {
 
   it("Should execute security updates", async () => {
     const securityUpdate = await shell(
-      `sudo bash ${hostScriptsPath}/security_update.sh`,
-      { noTimeout: true }
+      `sudo bash ${hostScriptsPath}/security_update.sh`
```
dapplion, line 31: Why is sudo necessary here? Could you add a comment explaining?

packages/dappmanager/test/integration/hostScripts.test.int.ts
```diff
@@ -25,4 +26,31 @@ describe("Host scripts", () => {
     );
     expect(hostInfo).to.be.ok;
   });
+
+  it("Should execute security updates", async () => {
+    try {
+      const securityUpdate = await shell(
+        `sudo bash ${hostScriptsPath}/security_update.sh`
```
dapplion, line 33: Can we find an alternative without sudo?

packages/dappmanager/test/integration/hostScripts.test.int.ts
pablomendezroyo: same as docker_update scripts. Will wrtie logs to a root path, where all dappnode stuff is:
```/usr/src/dappnode/logs```

I do not see it critical since will be run on github actions

We could solve this by creating an issue to only run these tests (critical tests that are susceptible to touch host machine) to be executed only on github actions

packages/dappmanager/test/integration/hostScripts.test.int.ts
pablomendezroyo: https://github.com/dappnode/DNP_DAPPMANAGER/pull/748#discussion_r625614512

PR #754, opened 2021-04-30
Added logs to docker update script

pablomendezroyo:


packages/dappmanager/hostScripts/docker_compose_update.sh
```diff
@@ -6,46 +6,60 @@ shopt -s nocasematch
...
+create_log_file
 
 function get_docker_compose_version() {
+  echo -e "\e[32mRetrieving compose version\e[0m" >> $LOG_FILE
   # Check if docker compose exists
   if type docker-compose >/dev/null 2>&1; then
     DOCKER_COMPOSE_VERSION=$(docker-compose version --short)
   else 
-    echo "docker compose does not exist or not recognized"
+    echo "docker compose does not exist or not recognized" 2>&1 | tee -a $LOG_FILE
```
3alpha, line 28: echo will always output only to stdout

pablomendezroyo: good point, fixed with last commit

packages/dappmanager/test/integration/hostScripts.test.int.ts
```diff
@@ -7,21 +7,21 @@ const hostScriptsPath = process.cwd() + "/hostScripts";
 describe("Host scripts", () => {
   it("Should fetch docker engine update requirements", async () => {
     const dockerEngineUpdateRequirements = await shell(
-      `bash ${hostScriptsPath}/docker_engine_update.sh --print-host-info`
+      `sudo bash ${hostScriptsPath}/docker_engine_update.sh --print-host-info`
```
dapplion, line 10: Why do you need sudo now?

pablomendezroyo: because it creates a log file at a reserved path: /usr/src/dappnode/logs

dapplion: Please find an alternative way to do that. Using sudo just for logs is not good

pablomendezroyo: https://github.com/dappnode/DNP_DAPPMANAGER/pull/748#discussion_r625614512

PR #756, opened 2021-05-03
Added logs migration

pablomendezroyo:


dapplion:
If I understand the purpose of this PR is to move some existing log files in the host to the new "central" location for logs.

If that's it, can we assume the existing logs as lost and care only about the ones we generate in the future? Not a fan of running a shellHost every time the dappmanager starts just for that niche possibility

pablomendezroyo:
> If I understand the purpose of this PR is to move some existing log files in the host to the new "central" location for logs.
> 
> If that's it, can we assume the existing logs as lost and care only about the ones we generate in the future? Not a fan of running a shellHost every time the dappmanager starts just for that niche possibility

This may be critical for the installer:
https://github.com/dappnode/DAppNode_Installer/blob/dfdeb6c575bb39e85cc94c001f83687043ca8202/build/dappnode/scripts/dappnode_install_pre.sh#L21

Not sure if there are any other places where these logs are used

Ideally, we should find a better way to check if logs were migrated or not with a less CPU cost and only once

Issue #760, opened 2021-05-05
Proxy IPFS avatars through DAPPMANAGER


dapplion:
The Admin UI has to be self-sufficient and not require additional services so it can be served through the HTTPs Portal

pablomendezroyo:
can be closed ?

PR #762, opened 2021-05-05
Add IPFS gateway proxy

dapplion:
Fixes https://github.com/dappnode/DNP_DAPPMANAGER/issues/760

Attaches a new HTTP proxy to the Ethfoward where requests with a pathname that starts with `ipfs-gateway-proxy` will be proxied to ipfs.dappnode:8080/ipfs

Re-uses the same code from the Ethforward, it's error handling and error views

---

I have not tested it yet on a real DAppNode

### DIFF
--- a/packages/dappmanager/src/ethForward/index.ts
+++ b/packages/dappmanager/src/ethForward/index.ts
@@ -1,137 +1,73 @@
-import http from "http";
-import httpProxy from "http-proxy";
 import express from "express";
 import params from "../params";
-import { EthProviderError } from "../modules/ethClient";
-import { ipfs } from "../modules/ipfs";
-import { urlJoin } from "../utils/url";
+import { getIpfsProxyHandler, ProxyType } from "./ipfsProxy";
 import { ResolveDomainWithCache } from "./resolveDomain";
-import * as views from "./views";
-import {
-  NodeNotAvailable,
-  ProxyError,
-  EnsResolverError,
-  NotFoundError,
-  Content
-} from "./types";
-import { logs } from "../logs";
 
 export function getEthForwardMiddleware(): express.RequestHandler {
-  const ethForwardHandler = getEthForwardHandler();
-  return (req, res, next): void => {
-    if (isDwebRequest(req)) ethForwardHandler(req, res);
-    else next();
-  };
-}
-
-/**
- * Convert a decentralized content into a fetchable URL
- * @param content
- */
-function getTargetUrl(content: Content): string {
-  switch (content.location) {
-    case "ipfs":
-      return urlJoin(params.ETHFORWARD_IPFS_REDIRECT, content.hash);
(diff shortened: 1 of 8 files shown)


packages/dappmanager/src/ethForward/index.ts
```diff
@@ -1,137 +1,79 @@
...
+  // Check if a request is for a decentralized website, based on their host
+  // - decentral.eth => true
+  // - my.dappmanager.dnp.dappnode.eth => false
+  // - my.dappnode => false
+  const domain = req.headers.host;
+  return typeof domain === "string" &&
+    domain.endsWith(".eth") &&
+    !domain.endsWith("dnp.dappnode.eth")
+    ? domain
+    : null;
```
pablomendezroyo, line 53: is return null the best approach for that case? wouldn't be better to create another function that returns a boolean if is a domain or not?

packages/dappmanager/src/ethForward/index.ts
```diff
@@ -1,137 +1,79 @@
...
+    async (_req, hash) => ({ location: "ipfs", hash })
+  );
 
-      // Must change the host from something not *.eth, to prevent the IPFS node
-      // from trying to resolve the domain with DNSLink and causing an error
-      req.headers.host = "dappmanager.dappnode";
+  return (req, res, next): void => {
+    try {
+      const domain = parseEthDomainHost(req);
+      if (domain !== null) {
```
pablomendezroyo, line 25: boolean function to check if is domain or not ?

packages/dappmanager/src/ethForward/index.ts
```diff
@@ -1,137 +1,79 @@
...
-        proxy.web(req, res, { target }, (e: Error | undefined, proxyRes) => {
-          if (!e) return resolve(proxyRes);
-          // Indicates that the host is unreachable.
-          // Usually happens when the node is not running
-          if (e.message.includes("EHOSTUNREACH"))
-            reject(new NodeNotAvailable(e.message, content.location));
-          else reject(new ProxyError(e.message, target));
-        });
-      });
+      const hash = parseIpfsGatewayProxyReqHash(req.url);
```
pablomendezroyo, line 30: boolean function to check if is hash or not ?

packages/dappmanager/src/ethForward/index.ts
dapplion: That's what we where doing before. The issue then is that you have to parse the domain twice. 1) to return true if it's a domain, 2) again to actually do the proxying

packages/dappmanager/src/ethForward/index.ts
dapplion: Same reasoning as with the domain. The goal is to not have to repeat work and only parse the hash once.

PR #770, opened 2021-05-14
HTTPs portal refresh button 

pablomendezroyo:
Containers on restart may loose it mapping in the https portal. A button to refresh the mapping should fix the issue

### DIFF
--- a/packages/admin-ui/src/pages/system/components/Network/HttpsMappings.tsx
+++ b/packages/admin-ui/src/pages/system/components/Network/HttpsMappings.tsx
@@ -189,8 +189,15 @@ export function HttpsMappings() {
             </React.Fragment>
           ))}
         </div>
-        <hr />
-        <Button onClick={() => refreshMapping}>Refresh</Button>
+
+        {mappings.data.length !== 0 && (
+          <>
+            <hr />
+            <Button onClick={() => refreshMapping(mappings.data || [])}>
+              Refresh
+            </Button>
+          </>
+        )}
       </>
     );
   }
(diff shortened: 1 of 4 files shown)


packages/admin-ui/src/pages/system/components/Network/HttpsMappings.tsx
```diff
@@ -16,13 +16,60 @@ import newTabProps from "utils/newTabProps";
...
+            httpsMapping.map(httpsMapping =>
+              api.httpsPortalMappingAdd(httpsMapping)
+            )
+          ),
+        {
+          message: "Adding HTTPs mapping...",
+          onSuccess: "Added HTTPs mapping"
+        }
+      );
+      setReqStatus({ result: true });
```
dapplion, line 65: This strategy is not safe. If the DAPPMANAGER <-> UI communications interrupt the "refresh mappings" operation will break at an unpredictable spot losing information. You should a new route called refreshMappings or something and do this atomically in the DAPPMANAGER side. And do it one be one:
```
for mapping:
  delete mapping
  add mapping
```

instead of 
```
for mapping:
  delete mapping
for mapping:
  add mapping
```

dapplion (changes requested):
See https://github.com/dappnode/DNP_DAPPMANAGER/pull/770/files#r632458559.

I would rename "Refresh" with "Recreate" it's more correct to what's happening.

On the visual side you should add a mention of what "Recreate" does and why it exists:
```
Use the "Recreate" mappings button if you experience issues.
```
- You should add a similar line to this in the main view
- You should add more though and longer text in the pop-up explaining again why this is used and what does it do. Current text in the popup is not sufficient
```
Refresh
Refresh HTTPs portal mapping
```

packages/admin-ui/src/pages/system/components/Network/HttpsMappings.tsx
```diff
@@ -25,43 +24,25 @@ export function HttpsMappings() {
...
-            )
-          ),
-        {
-          message: "Adding HTTPs mapping...",
-          onSuccess: "Added HTTPs mapping"
-        }
-      );
+      await withToast(() => api.httpsPortalMappingsRecreate(), {
+        message: "Recreating HTTPs mapping...",
+        onSuccess: "Recreated HTTPs mapping"
```
dapplion, line 43: mappings

dapplion (approved):

PR #772, opened 2021-05-16
support IPNS .eth domains

Leibniz137:


dapplion:
@Leibniz137 This looks great! Thank you so much for the contribution. Feel free to mark as "Read for review" when appropiate.

@pablomendezroyo Can you test this branch on a DAppNode and attempt to resolve `uniswap.eth`? Maybe we should add a unit test for it

dapplion:
@pablomendezroyo Bump

dapplion (dismissed):

dapplion:
@Leibniz137 I've been testing this approach and I can resolve uniswap.eth to `ipns-ns/12uA8M8Ku8mHUumxHcu7uee` but the IPFS gateway doesn't know what to do with it. Have you been able to resolve that result?

packages/dappmanager/test/ethForward/resolveDomain.test.int.ts
```diff
@@ -48,6 +55,14 @@ describe("ethForward > resolveDomain", () => {
     }
   });
 
+  describe("Fetch IPNS content", () => {
+    it.skip("Should resolve IPNS content", async () => {
+      // Resolve uniswap.eth
+      const res = await fetch(`${ipfsGateway}/ipns/12uA8M8Ku8mHUumxHcu7uee`);
```
Leibniz137, line 61: I just do `/ipns/uniswap.eth` (ie pass the name not the hash)

packages/dappmanager/src/ethForward/ipfsProxy.ts
```diff
@@ -88,18 +88,14 @@ export function getIpfsProxyHandler<T>(
...
-): string {
+function getTargetUrl(proxyType: ProxyType, content: Content): string {
   switch (proxyType) {
     case ProxyType.ETHFORWARD:
       switch (content.location) {
         case "ipfs":
           return urlJoin(params.ETHFORWARD_IPFS_REDIRECT, "ipfs", content.hash);
         case "ipns":
-          return urlJoin(params.ETHFORWARD_IPFS_REDIRECT, "ipns", domain);
+          return urlJoin(params.ETHFORWARD_IPFS_REDIRECT, "ipns", content.hash);
```
Leibniz137, line 98: here i think you actually want to keep passing the domain, we don't need to resolve it to the hash

PR #777, opened 2021-05-19
Wireguard multiple endpoints

pablomendezroyo:
Serve two configuration files:
- Remote
- Local

Todo: 
- [x] Find proper way to explain when to use each creds

Deppends on: https://github.com/dappnode/DNP_WIREGUARD/pull/10


dapplion (changes requested):
We should not show both the local and remote configs at once. There should be a naviation, button, switch or something that alternates between one on the other

dapplion (approved):
:fire:

3alpha (approved):

PR #793, opened 2021-05-31
Install/start avahi on host if necessary

pablomendezroyo:
We cannot ensure avahi-daemon is running and/or installed on existing dappnodes. It should be install/start if necessary

### DIFF
--- a/packages/dappmanager/src/daemons/avahi/index.ts
+++ b/packages/dappmanager/src/daemons/avahi/index.ts
@@ -3,10 +3,16 @@ import AbortController, { AbortSignal } from "abort-controller";
 import retry from "async-retry";
 import * as db from "../../db";
 import params from "../../params";
-import { pause } from "../../utils/asyncFlows";
 import { logs } from "../../logs";
 import { AvahiPublishCmdStatusType } from "../../types";
 import { LocalProxyingStatus } from "../../common";
+import {
+  installAvahiDaemon,
+  isAvahiDaemonInstalled,
+  isAvahiDaemonRunning,
+  startAvahiDaemonOnHost,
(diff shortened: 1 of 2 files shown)


packages/dappmanager/src/daemons/avahi/index.ts
```diff
@@ -35,6 +41,12 @@ class AvahiController {
         return;
       }
 
+      const isAvahiInstalled = await isAvahiDaemonInstalled();
+      if (!isAvahiInstalled) await installAvahiDaemon();
+
+      const isAvahiRunning = await isAvahiDaemonRunning();
+      if (!isAvahiRunning) await startAvahiDaemonOnHost();
```
dapplion, line 48: We should be able to refactor this into doing a single shellHost call

3alpha (approved):

packages/dappmanager/src/daemons/avahi/index.ts
dapplion: @pablomendezroyo The point of having a script was to not call shellHost 4 times but only 1

pablomendezroyo: I know, I just did the script, how and when to call it still need to be implemented

The idea would be:
- Call avahi_daemon.sh  --initialize on dappmanager startup, this would ensure:
    -  Install `avahi-tools` if necessary
    - Edit `/etc/hosts` and `/etc/avahi/avahi-daemon.conf` when hostname != dappnode. This situation might happen when install from script.
    - Start and enable avahi-daemon
- Use the other flags (`--start` `--stop` `--restart` `--status`) on demand by the user in the UI

PR #795, opened 2021-06-02
Avahi fixes

pablomendezroyo:
- Relocate avahi daemon startup
- Remove unnecesary revalidate

### DIFF
--- a/packages/admin-ui/src/params.ts
+++ b/packages/admin-ui/src/params.ts
@@ -38,7 +38,7 @@ export const wifiEnvWPA_PASSPHRASE = "WPA_PASSPHRASE";
 export const wifiEnvSSID = "SSID";
 
 // Local proxying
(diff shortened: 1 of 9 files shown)


packages/admin-ui/src/params.ts
```diff
@@ -38,7 +38,7 @@ export const wifiEnvWPA_PASSPHRASE = "WPA_PASSPHRASE";
 export const wifiEnvSSID = "SSID";
 
 // Local proxying
-export const adminUiLocalDomain = "my.dappnode.local";
+export const adminUiLocalDomain = "dappnode.local";
```
dapplion, line 41: Should this be `http://dappnode.local` and show the variable `adminUiLocalDomain` in the <a> above?

pablomendezroyo: Yes, nice catch. Done

dapplion (approved):

PR #798, opened 2021-06-03
Wireguard api refactor

pablomendezroyo:
Deppends on https://github.com/dappnode/DNP_WIREGUARD/pull/15/files

### DIFF
--- a/packages/dappmanager/src/modules/wireguard/client.ts
+++ b/packages/dappmanager/src/modules/wireguard/client.ts
@@ -11,8 +11,7 @@ const {
   WIREGUARD_DNP_NAME,
   WIREGUARD_ISCORE,
   WIREGUARD_MAIN_SERVICE,
-  WIREGUARD_LOCAL,
-  WIREGUARD_REMOTE
+  WIREGUARD_LOCAL
 } = params;
 
 export class WireguardClient {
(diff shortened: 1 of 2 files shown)


packages/dappmanager/src/modules/wireguard/client.ts
```diff
@@ -57,15 +56,19 @@ export class WireguardClient {
...
   ): Promise<WireguardDeviceCredentials> {
-    const remoteConfigUrl = urlJoin(
+    const remoteConfigUrl = urlJoin(WIREGUARD_API_URL, device);
+    const localConfigUrl = `${urlJoin(
       WIREGUARD_API_URL,
-      WIREGUARD_REMOTE,
       device
-    );
-    const localConfigUrl = urlJoin(WIREGUARD_API_URL, WIREGUARD_LOCAL, device);
+    )}?${WIREGUARD_LOCAL}`;
```
dapplion, line 71: This must not be a param, just use a string directly here.
```
const url = urlJoin(WIREGUARD_API_URL, device);
const remoteConfigUrl = url;
const localConfigUrl = `${url}?local=true`
```

pablomendezroyo: done on last commit

PR #800, opened 2021-06-04
Add .local alias

dapplion:
@3alpha @pablomendezroyo  Do we want to add this?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,5 +30,6 @@ services:
         aliases:
           - dappmanager.dappnode
           - my.dappnode
+          - dappnode.local
     logging:
       driver: journald
--- a/packages/dappmanager/src/domains.ts
+++ b/packages/dappmanager/src/domains.ts
@@ -47,7 +47,7 @@ export function getPrivateNetworkAliases(
 
   // Special unique alias for the Admin UI
   if (container.dnpName === params.dappmanagerDnpName)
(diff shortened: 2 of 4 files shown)


pablomendezroyo:
Im not sure what the alias `dappnode.local` would be used for.
As far as a I know it is only used by avahi-daemon.

dapplion:
I though we talked about this week. If a user connects via WIFI or VPN and navigates to this URL so that it still works.

pablomendezroyo:
> I though we talked about this week. If a user connects via WIFI or VPN and navigates to this URL so that it still works.

you right

pablomendezroyo (approved):

Jakic007:
This needs to be tested on linux machines because there are some configuration files which prevent resolution of .local domains via DNS servers after mDNS has failed.

dapplion:
> This needs to be tested on linux machines because there are some configuration files which prevent resolution of .local domains via DNS servers after mDNS has failed.

Right, that's definitely an issue to take into account. Tho having this on place is worse than having nothing?

3alpha:
It definitely won't break anything, but I am not sure it will work.

dapplion:
> It definitely won't break anything, but I am not sure it will work.

Do you think It's worth merging? @3alpha

3alpha:
I'll test it today and get back to you.

PR #826, opened 2021-06-25
Docs URLs links and other links

pablomendezroyo:
- Create component for DocsUrls links since are used in a lot of parts in the UI
- Open in new tab for external links in docs and community actions

### DIFF
--- a/packages/admin-ui/src/components/LinkDocs.tsx
+++ b/packages/admin-ui/src/components/LinkDocs.tsx
@@ -1,18 +1,17 @@
-import React from "react";
+import React, { AnchorHTMLAttributes } from "react";
 import newTabProps from "utils/newTabProps";
-import { FaReadme } from "react-icons/fa";
+import { BsInfoCircleFill } from "react-icons/bs";
 
-export default function LinkDocs({
-  urlDocs,
-  text
-}: {
-  urlDocs: string;
-  text?: string;
-}) {
+const LinkDocs: React.FC<AnchorHTMLAttributes<HTMLAnchorElement>> = ({
+  href,
+  children
+}) => {
   return (
-    <a style={{ display: "inLine" }} href={urlDocs} {...newTabProps}>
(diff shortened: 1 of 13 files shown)


packages/admin-ui/src/pages/community/components/DiscordActions.tsx
```diff
@@ -3,5 +3,7 @@ import React from "react";
 import { dappnodeDiscord } from "params";
 
 export default function DiscordActions() {
-  return <Button href={dappnodeDiscord}>Join Discord</Button>;
+  return (
+    <Button onClick={() => window.open(dappnodeDiscord)}>Join Discord</Button>
```
dapplion, line 7: Don't call window.open, just use native <a> components

packages/admin-ui/src/components/LinkDocs.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import newTabProps from "utils/newTabProps";
+import { FaReadme } from "react-icons/fa";
```
dapplion, line 3: Not a big fan of the FaReadme icon, looks like a butterfly to me haha. I would keep it simple and use an icon of info or help similar to (i) or (?)

packages/admin-ui/src/components/LinkDocs.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default function LinkDocs({
+  urlDocs,
+  text
+}: {
+  urlDocs: string;
+  text?: string;
+}) {
```
dapplion, line 11: You can simplify the API by renaming `urlDocs` to `href` and rendering children instead of text

```ts
<a style={{ display: "inLine" }} href={urlDocs} {...newTabProps}>
      <FaReadme className="links-icon" />
      {children}
    </a>
```

dapplion (approved):

PR #827, opened 2021-06-25
Check HTTPs is present in Local Network view

pablomendezroyo:


packages/admin-ui/src/pages/wifi/components/localProxying/LocalProxying.tsx
```diff
@@ -60,6 +64,25 @@ export function LocalProxying(): JSX.Element {
...
   }
 
+  // Helper UI in case the HTTPs Portal is bad
+  if (dnpsRequest.data) {
+    const httpsPortalDnp = dnpsRequest.data.find(
+      dnp => dnp.dnpName === httpsPortalDnpName
+    );
+    if (
+      !httpsPortalDnp &&
+      dappnodeIdentity.internalIp !== dappnodeIdentity.ip
```
dapplion, line 74: This check needs way more explanation, you must add a big comment here. Is it even necessary tho to do this check? Consider:
- Receiving a boolean from the DAPPMANAGER backend for `isBehindRouter` or something
- Always should the banner

packages/admin-ui/src/pages/wifi/components/localProxying/LocalProxying.tsx
```diff
@@ -124,15 +126,15 @@ export function LocalProxying(): JSX.Element {
           </div>
         </Card>
       ) : localProxyingStatus.isValidating ? (
-        <Loading steps={["Loading Local Network Proxy..."]} />
+        <Loading steps={["Loading Local Network Handler..."]} />
```
dapplion, line 130: What does it mean "handler" here?

packages/admin-ui/src/pages/wifi/components/localProxying/LocalProxying.tsx
```diff
@@ -7,19 +7,43 @@ import { ReqStatus } from "types";
...
 import { StateBadge } from "pages/packages/components/StateBadge";
 import { MdWifi } from "react-icons/md";
 import { parseContainerState } from "pages/packages/components/StateBadge/utils";
+import Alert from "react-bootstrap/esm/Alert";
 import { LocalProxyingStatus } from "common";
+import { NavLink } from "react-router-dom";
 import LinkDocs from "components/LinkDocs";
 
 export function LocalProxying(): JSX.Element {
+  const dnpsRequest = useApi.packagesGet();
```
dapplion, line 26: Just inline this logic in the same component. Now you won't show any help text describing this feature. So the text says "You must install the HTTPs Portal to use this feature", but what feature is it?

packages/admin-ui/src/pages/wifi/components/localProxying/LocalProxying.tsx
pablomendezroyo: back to proxy

packages/admin-ui/src/pages/wifi/components/localProxying/LocalProxying.tsx
pablomendezroyo: done

dapplion (approved):

PR #828, opened 2021-06-25
Improve UPnP available in initializeDB

pablomendezroyo:


packages/dappmanager/src/initializeDb.ts
```diff
@@ -71,7 +72,9 @@ export default async function initializeDb(): Promise<void> {
   //   UPnP is available and necessary only if the internalIp is not equal to the public IP
   //   and the external IP from UPnP command succeeded
   const upnpAvailable = publicIp
-    ? Boolean(externalIp && internalIp !== publicIp)
+    ? Boolean(externalIp && internalIp !== publicIp) && (await upnpc.list())
```
dapplion, line 75: Doesn't upnpc.list() throw if UPnP is not available?

pablomendezroyo: yes, and aren't this exactly what we want to set upnpavailable to false (if the function throws an error)

dapplion: Try it out, list() throwing will interrupt the entire initializeDb function and nothing will be set

pablomendezroyo: Then this error is a "controlled error" since we know the output and the meaning: upnp device is not available

In the file https://github.com/dappnode/DNP_DAPPMANAGER/blob/master/packages/dappmanager/src/modules/upnpc/parseGeneralErrors.ts

Should we change
```
if (noUpnpAvailable) throw Error("NOUPNP: No UPnP device available");
```

to 

```
if (noUpnpAvailable) {
   db.upnpavailable.set(true)
}
```

Or something similar

pablomendezroyo: upnp errors typed on last commit

packages/dappmanager/src/initializeDb.ts
```diff
@@ -71,11 +72,17 @@ export default async function initializeDb(): Promise<void> {
...
+  const upnpAvailable =
+    Boolean(publicIp && externalIp && internalIp !== publicIp) &&
+    upnpc
+      .list()
+      .then(() => true)
+      .catch(e => {
+        if (e instanceof UpnpError) return e.isUpnpAvailable;
+        return false;
+      })
       ? true
```
dapplion, line 84: This will always be true because you are not awaiting the promise. Create a helper function that returns `Promise<boolean>`, will make the code more readable

packages/dappmanager/src/modules/upnpc/upnpError.ts
```diff
@@ -1,45 +1,43 @@
...
-
-    // Check if it contains "failed"
-    return lastLine.includes("failed");
-  }
+// UPNPNOTAVAIL: UPnP is not available
+// UPNPNOTPORT: UPnP not existing port
+// UNKNOWN: UPnP unclassified error
+// Research for more errors!
+export enum UPnPErrors {
+  "UPNPNOTAVAIL",
```
dapplion, line 40: No need to use NodeJS garbage acronyms haha. You can use `UPNP_NOT_AVAILABLE`, space is free!

packages/dappmanager/src/modules/upnpc/upnpError.ts
```diff
@@ -1,45 +1,43 @@
...
+  "UPNPNOTPORT",
```
dapplion, line 41: If no-one uses this error type, no need to have it. We only care about the `UPNP_NOT_AVAILABLE` error

packages/dappmanager/src/modules/upnpc/upnpError.ts
```diff
@@ -1,45 +1,43 @@
...
-    terminalOutput,
-    command
-  }: {
-    terminalOutput: string;
-    command: string;
-  }) {
-    super(`UPnP error. Command failed: ${command}. Error: ${terminalOutput}`);
-    this.command = command;
+  constructor(terminalOutput: string, typeUpnpError: UPnPErrors) {
+    super(`UPnP Error: ${typeUpnpError}\nError:${terminalOutput}`);
```
dapplion, line 29: So know for all other errors we will get
```
UPnP Error: Unknown
Error: Bla bla bla bla
```
Not sure if that adds a lot. In that case would be better to just do
```
UPnP Error: Bla bla bla bla
```

packages/dappmanager/src/modules/upnpc/getExternalIp.ts
```diff
@@ -1,21 +1,18 @@
...
   try {
-    const res = await upnpcCommand(getExternalIpCommand);
+    const res = await upnpcCommand("-l");
     const externalIp = ((res || "").match(
       /ExternalIPAddress.=.((\d+\.?){4})/
     ) || [])[1];
     if (isIp(externalIp)) return externalIp;
     else throw Error("Wrong IP");
   } catch (e) {
+    const upnpError = parseUpnpErrors(e.message);
```
dapplion, line 14: The function signature is `parseUpnpErrors(terminalOutput: string)` however you are passing here the error message, which *should* contain the terminal output but it's prefixed with more things.

packages/dappmanager/src/initializeDb.ts
```diff
@@ -74,13 +73,7 @@ export default async function initializeDb(): Promise<void> {
   //   and the external IP from UPnP command succeeded
   const upnpAvailable =
     Boolean(publicIp && externalIp && internalIp !== publicIp) &&
```
dapplion, line 75: Do we use the `externalIp` for something? What does it even represent?

pablomendezroyo: we assume that when both IPs exist and are different from each other the DAppNode will not have a gateway with UPnP between itself and the internet

packages/dappmanager/src/modules/upnpc/upnpError.ts
pablomendezroyo: I would say could be useful for future improvements of UPnP. Once we know which actions to take depending on the error

dapplion (approved):

Issue #859, opened 2021-07-26
Add Prater dashboard


tropicar:
Prater package will be added soon, it would be nice that prater dashboard is visible.

tropicar:
This issue was fixed,

PR #860, opened 2021-07-26
Add prater dashboard

tropicar:
Add this feature #859


dapplion:
You don't need to harcode this data to the DAPPMANAGER, but add it in the manifest of the Prater package

PR #868, opened 2021-08-03
Skip compose up --no-recreate for packages containning pid

pablomendezroyo:
- Skip compose up --no-recreate for packages containning pid
- Added tests

### DIFF
--- a/packages/dappmanager/src/modules/installer/runPackages.ts
+++ b/packages/dappmanager/src/modules/installer/runPackages.ts
@@ -37,31 +37,33 @@ export async function runPackages(
     // - Allow copying files without duplicating logic
     // - Allow conditionally starting containers latter if were previously running
     log(pkg.dnpName, "Preparing package...");
-    await dockerComposeUp(pkg.composePath, {
-      // To clean-up changing multi-service packages, remove orphans
-      // but NOT for core packages, which always have orphans
-      removeOrphans: !pkg.isCore,
-      noStart: true,
-      timeout: pkg.dockerTimeout
-    });
+    if (!pkg.compose.services[pkg.dnpName].pid) {
+      await dockerComposeUp(pkg.composePath, {
+        // To clean-up changing multi-service packages, remove orphans
+        // but NOT for core packages, which always have orphans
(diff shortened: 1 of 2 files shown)


packages/dappmanager/src/modules/installer/runPackages.ts
```diff
@@ -76,3 +82,24 @@ export async function runPackages(
     log(pkg.dnpName, "Package started");
   }
 }
+
+/**
+ * Check if a package contains compose feature pid
+ * in any of its services
+ */
+export function hasPid(pkg: InstallPackageData): boolean {
```
dapplion, line 90: Can you move this to a utility function in compose? Rename to `composeHasPid()`. Also you can simplify with

```ts
for (const service of Object.values(pkg.compose.services) {
  if (service.pid) return true
}
return false
```

pablomendezroyo: done, double check since there is a change in the PR logic

packages/dappmanager/src/modules/installer/runPackages.ts
```diff
@@ -37,11 +38,15 @@ export async function runPackages(
     // - Allow copying files without duplicating logic
     // - Allow conditionally starting containers latter if were previously running
     log(pkg.dnpName, "Preparing package...");
+
+    // EXCEPTION!: If the compose contains: `pid:service.serviceName`
+    // compose must start with: `noStart: false`
```
dapplion, line 43: Add this comment right on top of the noStart property

packages/dappmanager/src/calls/packageRestart.ts
```diff
@@ -19,6 +21,8 @@ export async function packageRestart({
   if (!dnpName) throw Error("kwarg dnpName must be defined");
 
   const dnp = await listPackage({ dnpName });
+  const { compose } = new ComposeFileEditor(dnp.dnpName, dnp.isCore);
+  let isPidTargetServiceIncluded = false;
```
dapplion, line 25: Don't use let scoped variable unless there's not other way around it.

PR #874, opened 2021-08-19
fix typo in warning

pablomendezroyo:
use warning instead og warming


dapplion (approved):
Good pick! There is CI tooling that does english grammar and spell checking and source code files. Could you look into adding that into the DAPPMANAGER?

PR #885, opened 2021-08-24
Set dyndns on dappmanager startup faster

pablomendezroyo:
Dyndns is needed for other services to be exposed on the dappmanager API as soon as possible. This process may take a little time on dappmanager startup due to other processes such as getting IPs or other daemons.

- Get dynds on dappmanager startup
- Set dynds on db before start dappmanager api


dapplion:
I don't see how the PR fixes the issue. What you need to do is run generateKeys() on startup

pablomendezroyo:
not needed

PR #908, opened 2021-10-11
LVM: expand/reduce host filesystem space

pablomendezroyo:
Expand the disk space of the dappnode with LVM

screenshots
image
image
image


3alpha:
We need to really think over warning messages we show users before doing this since this is critical operation. 

Also, there is still pending research to be done. This comes to mind:

- What happens if you add slower drive?
- What happens if you disconnect drive before releasing it from the LVM?

pablomendezroyo:
> > Is a regular user able to take this decisions without external help? The UI should link to well written resources in the docs AND have a minimal guide of why each of the option should be chosen. Also the DAPPMANAGER should try to guess a sane default in case the user just goes "next", "next", "next"
> 
> We definitely need docs which would describe every option and advise them of "default" ones, but I would disallow users going "next", "next" and make them fill out fields themselves because that will make them think.

I was thinking on writting a pretty and well documented **how-to** in the forum explaining how to upgrade the disk space with screenshots and pictures

PR #918, opened 2021-11-05
fix bug in dataSend

pablomendezroyo:
Prevent from crashing the `dataSend` feature

### DIFF
--- a/packages/dappmanager/src/api/routes/dataSend.ts
+++ b/packages/dappmanager/src/api/routes/dataSend.ts
@@ -11,11 +11,12 @@ const MAX_KEYS = 20;
  */
 export const dataSend = wrapHandler(async (req, res) => {
   const key = req.query.key;
-  const data = req.query.data || req.body;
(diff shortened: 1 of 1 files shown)


dapplion:
If I can guess you run this command
```
curl -X POST http://my.dappnode/data-send?key=apiKey&data=AAXXAAXXAAXXAAXXAAXXAAXXAAXXAAXXAAXXAAXX
```
The problem is that bash interprets the string's `&` character not a raw string and omits the rest so the server receives
```
curl -X POST http://my.dappnode/data-send?key=apiKey
```
Then the validation function throws because the `data` param is undefined.

To send multiple URL params you must quote the URL like
```
curl -X POST "http://my.dappnode/data-send?key=apiKey&data=AAXXAAXXAAXXAAXXAAXXAAXXAAXXAAXXAAXXAAXX"
```

pablomendezroyo:
PR not needed, curl must be done with double quotes

PR #922, opened 2021-11-17
Render link in package sent data

pablomendezroyo:
If the data sent from the container to the dappmanager starts with **http** then render it as a Link. Related to https://github.com/dappnode/DAppNodePackage-prysm-prater/issues/19

Consider having a hardcoded key instead for links/URLs

### DIFF
--- a/packages/admin-ui/src/pages/packages/components/Info/PackageSentData.tsx
+++ b/packages/admin-ui/src/pages/packages/components/Info/PackageSentData.tsx
@@ -72,7 +72,7 @@ function SentDataRow({
   isSecret
 }: {
   value: string;
-  isLink: boolean;
+  isLink?: boolean;
   isSecret?: boolean;
 }) {
   const [show, setShow] = useState(false);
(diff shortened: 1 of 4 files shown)


dapplion:
Please add styles to the link box so that it looks consistent with the input. Also you must set proper word-break so it doesn't break the layout
```css
padding: .375rem .75rem;
border: 1px solid #ced4da;
border-radius: .25rem;
word-break: break-all;
```
Screenshot from 2021-11-17 10-47-56

packages/admin-ui/src/pages/packages/components/Info/PackageSentData.tsx
```diff
@@ -61,9 +68,11 @@ export function RenderPackageSentData({
 
 function SentDataRow({
   value,
+  isLink,
   isSecret
 }: {
   value: string;
+  isLink: boolean;
```
dapplion, line 75: Make optional

packages/admin-ui/src/utils/isLink.ts
```diff
@@ -0,0 +1,3 @@
+export function isLink(value: string): boolean {
+  return value.startsWith("http");
```
dapplion, line 2: This condition should be more resilient:
```ts
return value.startsWith("http://") || value.startsWith("https://");
```

PR #926, opened 2021-11-23
Fix link in dataSent feature

pablomendezroyo:
There was used a link with the full url path, instead it should be used only the package ENS domain


dapplion (commented):
Looks good!

PR #931, opened 2021-11-30
Fix get manifest from IPFS remote

pablomendezroyo:
The IPFS method `dag/export/ does not allow to use IPFS hashes in path format, i.e:
`QmXcZnd6euc3wfQqwb8m9r2hWuefrS6cPic29k5VZz7S6d/dappnode_package.json`

There must be used the CID of the file


dapplion (approved):
Looks good!

PR #936, opened 2021-12-15
eth2 migration

pablomendezroyo:
Eth2migration from Prysm pre-web3signer to Eth2 client with web3signer

- [x] Implement import
- [x] Implement export
- [x] Implement integration test
- [ ] Implement in the eth2 client side (prysm, teku, lighthouse) entrypoint to fetch pubkeys from web3signer API `list` method
    - [x] Lighthouse: :heavy_check_mark: 
    - [x] Teku: :heavy_check_mark: 
    - [ ] Prysm: :question:  to be implemented in Prysm
- [x] Implement frontend for the migration


packages/dappmanager/src/modules/eth2migration/export/exportKeys.ts
```diff
@@ -0,0 +1,48 @@
...
+    // Get public keys in a string comma separated
+    const validatorAccountsDataParsed = parseValidatorAccounts(
+      validatorAccountsData
+    );
+
+    // 2. Copy walletpassowrd to backup folder
+    shell(
+      `docker exec ${containerName} cp ${eth2migrationParams.keys.walletPasswordFile} ${eth2migrationParams.backup.backupDir}`
+    ).catch(e => {
+      throw Error(`Eth2 migration: backup wallet password failed. ${e}`);
```
dapplion, line 30: Don't interpolate Error objects in a template literal. Extend the error message and re-throw

packages/dappmanager/src/modules/eth2migration/getPrysmValidatorImage.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export async function getPrysmValidatorImage({
+  prysmOldDnpName,
+  network
+}: {
+  prysmOldDnpName: string;
+  network: Eth2Network;
+}): Promise<string> {
+  // TODO: determine stable versions that support the prysm cli used in the eth2migration
+  const prysmStableVersion = network === "mainnet" ? "1.0.22" : "0.1.7";
```
dapplion: Don't assume there are only two networks, there may be more than mainnet and prater

pablomendezroyo: agree, now is retrieved from the starting function `getMigrationParams` depending on the network

pablomendezroyo:
not needed anymore

PR #947, opened 2022-01-29
Move away from non RFC1918 compliant Network Address

lodotek:


dapplion:
@3alpha Please can you review?

3alpha:
It is not going to work like this, because it would break everything. Migration has to be system wide, not just in DAppManager.

lodotek:
Yes of course it will need to be changed everywhere.

3alpha:
It is not just about changing, but also about migration since all packages need to be updated and that happens in asynchronous fashion. I am closing this PR now and I will open an issue which will deal with this in multistep approach.

PR #953, opened 2022-02-04
deduplicate ports with ipv4 and ipv6

pablomendezroyo:
### BUG

The docker api returns two responses for the same port request: one for the ipv4 and the other one for the ipv6 (if available)

This situation was not contemplated in the backend and was returning fake duplicated ports

### FIX 
- Deduplicate ports for ipv4 and ipv6
- Add new type when the port has ipv6

### DIFF
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -371,7 +371,6 @@ export interface PortMapping extends BasicPortMapping {
   ephemeral?: boolean;
   ip?: string;
   deletable?: boolean;
-  hasIpv6?: boolean;
 }
 
 export interface PackagePort {
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -371,6 +371,7 @@ export interface PortMapping extends BasicPortMapping {
   ephemeral?: boolean;
   ip?: string;
   deletable?: boolean;
+  hasIpv6?: boolean;
 }
 
 export interface PackagePort {
--- a/packages/dappmanager/src/modules/docker/utils.ts
+++ b/packages/dappmanager/src/modules/docker/utils.ts
@@ -98,11 +98,11 @@ export function getDockerTimeoutMax(
     ]
  * ```
  */
-export function parseDockerApiListPorts(
+export function ensureUniquePortsFromDockerApi(
   ports: Dockerode.Port[],
   defaultPorts: PortMapping[] | undefined
 ): PortMapping[] {
-  return ports
+  const portsFromApi = ports
     .filter(port => port.IP && port.IP.trim() !== "::")
     .map(
       ({ PrivatePort, PublicPort, Type }): PortMapping =>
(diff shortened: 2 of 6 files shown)


packages/dappmanager/src/modules/docker/utils.ts
```diff
@@ -68,3 +71,59 @@ export function getDockerTimeoutMax(
...
+      {
+        "IP": "::",
+        "PrivatePort": 8000,
+        "PublicPort": 2345,
+        "Type": "tcp"
+      }
+    ]
+ * ```
+ */
+export function parseDockerApiListPorts(
```
dapplion, line 101: This does not de-duplicate ports. You must ensure that the UI condition of one port per protocol and container port.

packages/admin-ui/src/common/types.ts
```diff
@@ -371,6 +371,7 @@ export interface PortMapping extends BasicPortMapping {
   ephemeral?: boolean;
   ip?: string;
   deletable?: boolean;
+  hasIpv6?: boolean;
```
dapplion, line 374: If there is no-one consuming this property remove. We should add features only when useful

dapplion (changes requested):

packages/admin-ui/src/common/types.ts
pablomendezroyo: I know, but who knows in the future?

packages/dappmanager/src/modules/docker/utils.ts
pablomendezroyo: right! done

packages/admin-ui/src/common/types.ts
pablomendezroyo: removed type

dapplion: > I know, but who knows in the future?

Then we'll add it in the future when necessary

packages/dappmanager/src/modules/docker/utils.ts
```diff
@@ -68,3 +71,55 @@ export function getDockerTimeoutMax(
...
+    )
+    .map(
+      (port): PortMapping => ({
+        ...port,
+        deletable: isPortMappingDeletable(port, defaultPorts)
+      })
+    );
+
+  // deduplicate ports
+  return [...new Set(portsFromApi)];
```
dapplion, line 124: This does not de-duplicate ports, since different unique objects can represent the same pair of container port and protocol. You must de-duplicate by the key (container port + protocol) as defined in the UI

packages/dappmanager/src/modules/docker/utils.ts
```diff
@@ -68,3 +71,55 @@ export function getDockerTimeoutMax(
...
+      }
+    ]
+ * ```
+ */
+export function ensureUniquePortsFromDockerApi(
+  ports: Dockerode.Port[],
+  defaultPorts: PortMapping[] | undefined
+): PortMapping[] {
+  const portsFromApi = ports
+    .filter(port => port.IP && port.IP.trim() !== "::")
```
dapplion, line 106: Remove, don't make any assumptions on how the IP is rendered, de-duplication must be done per container port and protocol

dapplion (changes requested):
See notes

packages/dappmanager/src/modules/docker/utils.ts
pablomendezroyo: fixed!

dapplion (approved):

lodotek:
@dapplion @pablomendezroyo how can we get this to take effect on our existing Dappnode installations?

alexpeterson91:
When the next DAPPMANAGER version is published it will be updatable via the UI.  I do not know how soon the next release will be since we try to do updates to core packages only when necessary and there’s a few other major outstanding issues that we may try to finish and roll it all into the next version. 

Currently It can be built using the SDK and installed manually or via the relevant IPFS hash available on GitHub, if you know how but it’s not recommended unless you really know what you’re doing and do it at your own risk

PR #955, opened 2022-02-07
Refactor drivers module beaconchain API

pablomendezroyo:
Fetch the standard API endpoint to calculate beaonchain sync status
https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus


packages/admin-ui/src/common/types.ts
```diff
@@ -429,15 +429,11 @@ export type ContainerState =
...
-export type ChainDriver =
-  | "bitcoin"
-  | "ethereum"
-  | "ethereum2-beacon-chain-prysm"
-  | "monero";
+export type ChainDriver = "bitcoin" | "ethereum" | "ethereum2" | "monero";
 export const chainDrivers: ChainDriver[] = [
   "bitcoin",
   "ethereum",
-  "ethereum2-beacon-chain-prysm",
```
dapplion, line 440: Consider keeping this for backwards compatibility, don't break current versions of Prysm

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,130 @@
...
+      syncing: false,
+      message: "No node syncing data",
+      error: true,
+      progress: 0
+    };
+
+  const { head_slot, sync_distance, is_syncing } = nodeSyncing.data;
+  const highestBlock = parseInt(head_slot) + parseInt(sync_distance);
+
+  if (highestBlock - parseInt(head_slot) < MIN_SLOT_DIFF_SYNC) {
```
dapplion, line 69: Why do the math again? you already have `sync_distance` this line is
```
parseInt(head_slot) + parseInt(sync_distance) - parseInt(head_slot) < MIN_SLOT_DIFF_SYNC
```
which is equivalent to
```
parseInt(sync_distance) < MIN_SLOT_DIFF_SYNC
```

dapplion: Add a helper to parse numbers like

```ts
function toNum(numStr: string): number {
  const num = parseInt(numStr, 10)
  if (isNaN(num)) throw Error(`${numStr} is not a number`)
  return num
}
```

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,130 @@
...
+    return {
+      syncing: false,
+      error: false,
+      message: `Synced #${head_slot}`
+    };
+  } else {
+    return {
+      syncing: is_syncing,
+      message: `Blocks synced ${head_slot} / ${highestBlock}`,
+      progress: safeProgress(parseInt(head_slot) / highestBlock),
```
dapplion, line 79: Parse everything once at the top with a safe function

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  if (highestBlock - parseInt(head_slot) < MIN_SLOT_DIFF_SYNC) {
+    return {
+      syncing: false,
+      error: false,
+      message: `Synced #${head_slot}`
+    };
+  } else {
+    return {
+      syncing: is_syncing,
```
dapplion, line 77: Should you use `is_syncing` here? You are deciding if it's synced or not with `MIN_SLOT_DIFF_SYNC`, just set to true

packages/admin-ui/src/common/types.ts
pablomendezroyo: it is compatible with current prysm version, so we could just name it ethereum2

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
pablomendezroyo: why say it is syncing whereas it's not?  there can be up to four different states:
- syncing
- synced
- not syncing + error
- syncing but no error

This would be the fourth case

dapplion:
> it is compatible with current prysm version, so we could just replace it with the new one

Are you sure? You can't change the manifest of already installed packages

I think this will throw error "unknown driver eth2-prysm"

pablomendezroyo:
> > it is compatible with current prysm version, so we could just replace it with the new one
> 
> Are you sure? You can't change the manifest of already installed packages
> 
> I think this will throw error "unknown driver eth2-prysm"

oh now I get you, then I will add ` eth2-prysm | ethereum2`

packages/dappmanager/src/modules/chains/drivers/index.ts
```diff
@@ -25,8 +25,8 @@ export async function runWithChainDriver(
     case "ethereum":
       return await ethereum(dnp);
 
-    case "ethereum2-beacon-chain-prysm":
-      return await ethereum2Prysm(dnp);
+    case "ethereum2" || "ethereum2-beacon-chain-prysm":
```
dapplion, line 28: Is this valid syntax? I think here you are testing only `"ethereum2"``can you unit test a case for `"ethereum2-beacon-chain-prysm"`?

pablomendezroyo: fixed or

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
...
+import { safeProgress } from "../utils";
+
+const beaconChainServiceName = "beacon-chain";
+const MIN_SLOT_DIFF_SYNC = 60;
+
+// Wait for Promises to resolve. Do not cache rejections
+// Cache for 1 hour, genesis and config should never change
+const fetchNodeSyncingMemo = memoize(fetchNodeSyncingStatus, {
+  promise: true,
+  maxAge: 3e6
```
dapplion, line 16: You MUST not cache the syncing result for 1 hour, it changes almost every second. This caching was here exclusively for genesis and config calls, which never change.

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
...
+    throw Error(`${beaconChainServiceName} service not found`);
+  }
+  if (!beaconChainContainer.running) {
+    return null; // OK to not be running, just ignore
+  }
+
+  const containerDomain = getPrivateNetworkAlias(beaconChainContainer);
+
+  // http://beacon-chain.prysm-pyrmont.dappnode:3500/
+  const apiUrl = `http://${containerDomain}:3500`;
```
dapplion, line 39: Warning! Each client has it's own port. 3500 only applies to Prysm. Each client should declare the port in the manifest

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
...
+      message: `Could not connect to RPC. ${e.message}`,
+      error: true
+    };
+  }
+}
+
+/**
+ * Parses the response from the beacon node to describe if it's currently syncing or not, and if it is, what block it is up to.
+ */
+export function parseEthereum2State(nodeSyncing: NodeSyncing): ChainDataResult {
```
dapplion, line 58: You are not parsing the state, rename to `parseNodeSyncingResponse`

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
...
+ */
+async function fetchNodeSyncingStatus(baseUrl: string): Promise<NodeSyncing> {
+  const response: NodeSyncing = await fetchPrysmApi(
+    baseUrl,
+    "/eth/v1/node/syncing"
+  );
+  return response;
+}
+
+async function fetchPrysmApi<T>(baseUrl: string, apiPath: string): Promise<T> {
```
dapplion, line 104: Merge `fetchPrysmApi` with `fetchNodeSyncingStatus`, having two functions is unnecessary and this driver does not only apply to Prysm

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
...
+    };
+  }
+}
+
+/**
+ * Requests the beacon node to describe if it's currently syncing or not, and if it is, what block it is up to.
+ * https://ethereum.github.io/beacon-APIs/#/Node/getSyncingStatus
+ */
+async function fetchNodeSyncingStatus(baseUrl: string): Promise<NodeSyncing> {
+  const response: NodeSyncing = await fetchPrysmApi(
```
dapplion, line 97: You can return directly here

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
...
+      progress: 0
+    };
+
+  const { head_slot, sync_distance, is_syncing } = nodeSyncing.data;
+  const headSlot = toNum(head_slot);
+  const syncDistance = toNum(sync_distance);
+  const highestBlock = headSlot + syncDistance;
+  const progress = safeProgress(headSlot / highestBlock);
+
+  if (highestBlock - headSlot < MIN_SLOT_DIFF_SYNC) {
```
dapplion, line 74:
```ts
if (syncDistance < MIN_SLOT_DIFF_SYNC) {
```

dapplion (changes requested):

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
```diff
@@ -0,0 +1,145 @@
+import fetch from "node-fetch";
+import memoize from "memoizee";
+import { getPrivateNetworkAlias } from "../../../domains";
+import { urlJoin } from "../../../utils/url";
+import { InstalledPackageData } from "../../../types";
+import { ChainDataResult } from "../types";
+import { safeProgress } from "../utils";
+
+const beaconChainServiceName = "beacon-chain";
```
dapplion, line 9: You can't ensure that all clients will have the container with this name. This should be a configurable parameter via the manifest that defaults to `beacon-chain` only if not specified

pablomendezroyo: implement on https://github.com/dappnode/DNP_DAPPMANAGER/pull/963

packages/dappmanager/src/modules/chains/drivers/ethereum2.ts
pablomendezroyo: implement on https://github.com/dappnode/DNP_DAPPMANAGER/pull/963

dapplion (approved):
LGTM!

PR #959, opened 2022-02-10
Switch from security updates to full upgrade

pablomendezroyo:
The security updates was giving faiure in Buster, also it was requested by many users the ability to do a fill apt `update && apt upgrade` directly from the UI

Consider:
- Relocate the UI script execution location
- Any other suggestion for the `apt-upgrade` command?

cc: @3alpha


3alpha (commented):
Should this script be renamed to reflect its purpose?

pablomendezroyo:
> Should this script be renamed to reflect its purpose?

good point. Yes, it should be renamed

pablomendezroyo:
> I am having second thoughts about this, maybe we should go this route: https://wiki.debian.org/UnattendedUpgrades

I would consider automatic updates danger for the user, could they break dappnode host in some edge cases? maybe could be something to be automated on demand by the user

3alpha:
We could think about it and we can merge this for now. Only we need to change UI displays and warning as well.

dapplion (approved):

PR #965, opened 2022-02-16
Adding external network to DNP_HTTPS compose file on first mapping

### LINKED ISSUE #964: Ensure that HTTPS portal is always connected to the external network
## Issue
DAppManager never changes Compose file of DNP_HTTPS package so when container restarts, connection is lost.

## Solution
When first mapping is created, make sure that DNP_HTTPS compose is changed.

3alpha:
Resolves #964

### DIFF
--- a/packages/dappmanager/src/modules/https-portal/index.ts
+++ b/packages/dappmanager/src/modules/https-portal/index.ts
@@ -10,6 +10,7 @@ import { getExternalNetworkAlias } from "../../domains";
 import { PackageContainer, HttpsPortalMapping } from "../../types";
 import { HttpsPortalApiClient } from "./apiClient";
 import { addNetworkAliasCompose, removeNetworkAliasCompose } from "./utils";
+import { ComposeEditor } from "../compose/editor";
 export { addAliasToRunningContainersMigration } from "./migration";
 export { HttpsPortalApiClient };
 export { getExposableServices } from "./exposable";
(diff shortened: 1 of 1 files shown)


packages/dappmanager/src/modules/https-portal/index.ts
```diff
@@ -67,6 +68,16 @@ export class HttpsPortal {
 
     // Edit compose to persist the setting
     addNetworkAliasCompose(container, externalNetworkName, aliases);
+
+
+    // Check whether DNP_HTTPS compose has external network persisted
+    const httpsComposePath = ComposeEditor.getComposePath(params.HTTPS_PORTAL_DNPNAME, true)
+    const editor = new ComposeEditor(ComposeEditor.readFrom(httpsComposePath))
+
+    if(editor.getComposeNetwork(externalNetworkName) === null) {
```
pablomendezroyo, line 77: Since null is a nullish value I would use a nullish check for all possible nullish values
```
if(editor.getComposeNetwork(externalNetworkName) === null)
```

pablomendezroyo:
LGTM! It could be implemented into a utility function that could be unit tested, but I guess its overkilled

packages/dappmanager/src/modules/https-portal/index.ts
3alpha: Only nullish value `editor.getComposeNetwork` can return is `null`. But, I can change it if improves readability.

pablomendezroyo: an empty string is a nullish value as well

pablomendezroyo (approved):

packages/dappmanager/src/modules/https-portal/index.ts
3alpha: That is not relevant since:
```
  getComposeNetwork(networkName: string): ComposeNetwork | null {
    return this.compose.networks?.[networkName] ?? null;
  }
  ```

File: packages/dappmanager/src/modules/https-portal/index.ts [hunk 2]
```diff
@@ -67,6 +68,16 @@ export class HttpsPortal {
...
+      const httpsExternalAlias = getExternalNetworkAlias(httpsPortalContainer)
+      addNetworkAliasCompose(httpsPortalContainer, externalNetworkName, [httpsExternalAlias])
```
dapplion, line 79: The compose has to be `docker-compose up` to take effect

dapplion (changes requested):

packages/dappmanager/src/modules/https-portal/index.ts
3alpha: To take effect as in change compose file or as in connect to the network? If it is the latter we already take care of that few lines up.

dapplion: Oh right, should be okay then

dapplion (approved):
LGMT

PR #966, opened 2022-02-17
Track package registries in xDAI

dapplion:
See https://github.com/dappnode/DAppNode/issues/327 for full rationale

Changes in DAPPMANAGER:
- Merge fetchDirectory and fetchRegistry into the same action.
  - fetchDirectory iterated a list of packages in the directory contract. Now that information is available through state calls in the registry contract
  - fetchRegistry scanned events on the registry contract to discover packages. Now that information is available through state calls in the registry contract
- Resolve names through local table of registryAddresses + registry contract state calls. Previously we used ENS
- `isVerified` status now comes from the registry contract itself instead of the dnpName suffix

### TODO
- [x] Update node URLs to point to xDAI node instead of mainnet
- [x] Update repository source options
- [x] Add integration tests

### Changes

**Repository source**

Since now it tracks xDAI, there is only one node option, which does not support light-clients.

Screenshot from 2022-04-04 14-34-15

### Future

- [ ]  How to expand to arbitrary registries, provided by users and developers
- [ ]  How to have registries in multiple chains
- [ ]  How to resolve registry collisions
- [ ]  How to resolve repo name collisions


dapplion:
Deployed a version of the registry with a burner account here https://blockscout.com/xdai/mainnet/address/0x01c58A553F92A61Fd713e6006fa7D1d82044c389

This is the registry address used in this PR, but should be replaced by a release with a long term account

PR #972, opened 2022-02-25
Expose https ports by default on installation

pablomendezroyo:
Expose HTTPS ports defined in the manifest with `exposedByDefault` as `true` 

e.g
``` 
exposable: {
    "name": "Geth JSON RPC",
    "description": "JSON RPC endpoint for Geth mainnet",
    "serviceName": "beacon_chain",
    "port":  8545,
    "exposedByDefault": true
}
```

### DIFF
--- a/packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
+++ b/packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
@@ -1,0 +1,78 @@
+import { listPackageNoThrow } from "../docker/list/listPackages";
+import { httpsPortal } from "../../calls/httpsPortal";
+import { prettyDnpName } from "../../utils/format";
+import params from "../../params";
+import { InstallPackageData } from "../../types";
+import { Log } from "../../utils/logUi";
+import { HttpsPortalMapping } from "../../common";
+
+/**
+ * Expose default HTTPS ports on installation defined in the manifest - exposable
+ */
+export async function exposeByDefaultHttpsPorts(
+  pkg: InstallPackageData,
+  log: Log
+): Promise<void> {
+  if (pkg.metadata.exposable) {
+    const portMappinRollback: HttpsPortalMapping[] = [];
+    for (const exposable of pkg.metadata.exposable) {
+      if (exposable.exposedByDefault) {
+        // Check HTTPS package exists
+        const httpsPackage = await listPackageNoThrow({
+          dnpName: params.HTTPS_PORTAL_DNPNAME
+        });
+        if (!httpsPackage)
+          throw Error(
+            `HTTPS package not found but required to expose HTTPS ports by default. Install HTTPS package first.`
+          );
+        // Check HTTPS package running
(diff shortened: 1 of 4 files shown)


packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
```diff
@@ -0,0 +1,78 @@
...
+        try {
+          // Expose default HTTPS ports
+          log(
+            pkg.dnpName,
+            `Exposing ${prettyDnpName(pkg.dnpName)}:${
+              exposable.port
+            } to the external internet`
+          );
+          await httpsPortal.addMapping(portalMapping);
+          portMappinRollback.push(portalMapping);
```
dapplion, line 53: Why is a rollback mechanism necessary here?

packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
```diff
@@ -0,0 +1,78 @@
...
+          log(
+            pkg.dnpName,
+            `Exposed ${prettyDnpName(pkg.dnpName)}:${
+              exposable.port
+            } to the external internet`
+          );
+        } catch (e) {
+          e.message = `${e.message} Error exposing default HTTPS ports, removing mappings`;
+          for (const mappingRollback of portMappinRollback) {
+            await httpsPortal.removeMapping(mappingRollback).catch(e => {
```
dapplion, line 64: Just rollback `portalMapping` not all of them

packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Expose default HTTPS ports on installation defined in the manifest - exposable
+ */
+export async function exposeByDefaultHttpsPorts(
+  pkg: InstallPackageData,
+  log: Log
+): Promise<void> {
+  if (pkg.metadata.exposable) {
+    const portMappinRollback: HttpsPortalMapping[] = [];
+    for (const exposable of pkg.metadata.exposable) {
```
dapplion, line 18:
```ts
const exposableArr = pkg.metadata.exposable.filter(
  exposable => exposable.exposeByDefault
)

if (exposableArr.length > 0) {
  // Check HTTPS package exists once
]


```

packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
dapplion: Offline chat: To prevent having exposed ports with a non-installed package

packages/dappmanager/src/modules/installer/exposeByDefaultHttpsPorts.ts
dapplion: No, we should rollback all previously open

dapplion (approved):
LGTM!

PR #974, opened 2022-03-01
Disable updated button

pablomendezroyo:
Disable button if `isUpdated`

### DIFF
--- a/packages/admin-ui/src/pages/installer/components/Steps/Info.tsx
+++ b/packages/admin-ui/src/pages/installer/components/Steps/Info.tsx
@@ -196,7 +196,7 @@ export const InstallerStepInfo: React.FC<InstallerStepInfoProps> = ({
                   className="install"
                   variant="dappnode"
                   onClick={onInstall}
-                  disabled={disableInstallation}
+                  disabled={disableInstallation || isUpdated}
                 >
                   {tagDisplay}
                 </Button>


3alpha (approved):

dapplion (approved):

dapplion:
> I have one comment here. I agree this is confusing to allow the button to be clicked but is there another way to start the setup_wizard from the GUI? Or have all fields been made available in the config tab? I know at one point months back I had to go through the updated button to change profile settings on IPFS but that hasn't been the case in a while now. Just wanna make sure we don't lock ourselves out of a setup wizard.

For installed packages you can see the setup wizard at each package config package. For IPFS, got to UI -> packages -> IPFS -> config (tab)

Can you double check if you can get to the page you expect that way?

> Also question does the version checking work by checking the package IPFS hashes or does it compare the version data?

AFAIK it compares the package semver with the installed semver

alexpeterson91:
@dapplion i dont know if SemVer will work for this, if the semver of a package is higher than currently released it assumes you are updated and running the latest package, even if you're not, downgrading to the latest release wont be possible.  Im not totally sure how each function checks for updated or not but i have found some irregularities with a couple packages on my test machines, where this being released could cause me issues id need to go to GH or the package explorer to grab a hash of the latest release or whatever.  i dont know enough about it but i think this deserves another look and review.

dapplion:
> @dapplion i dont know if SemVer will work for this, if the semver of a package is higher than currently released it assumes you are updated and running the latest package, even if you're not, downgrading to the latest release wont be possible from the UI as is is currently. Im not totally sure how each function checks for updated or not but i have found some irregularities with a couple packages on my test machines, where this being released could cause me issues id need to go to GH or the package explorer to grab a hash of the latest release or whatever. i dont know enough about it but i think this deserves another look and review.

Definitely would need a more in depth review, @alexpeterson91 @pablomendezroyo can you open an issue?

pablomendezroyo:
> > @dapplion i dont know if SemVer will work for this, if the semver of a package is higher than currently released it assumes you are updated and running the latest package, even if you're not, downgrading to the latest release wont be possible from the UI as is is currently. Im not totally sure how each function checks for updated or not but i have found some irregularities with a couple packages on my test machines, where this being released could cause me issues id need to go to GH or the package explorer to grab a hash of the latest release or whatever. i dont know enough about it but i think this deserves another look and review.
> 
> Definitely would need a more in depth review, @alexpeterson91 @pablomendezroyo can you open an issue?

@alexpeterson91 not sure what Im looking for, feel free to open an issue with a very specific detailed problem

PR #980, opened 2022-03-07
Ensure not installing legacy validators

pablomendezroyo:
The dappstore will not allow to install older versions of package however users still can use hashes to install validator package previous to the remote signer support.

This PR checks a hardcoded version for the Prysm prater to ensure it supports remote signer. Prysm mainnet and Gnosis is on the roadmap

### DIFF
--- a/packages/dappmanager/src/modules/installer/ensureIsNotValidatorLegacyPackage.ts
+++ b/packages/dappmanager/src/modules/installer/ensureIsNotValidatorLegacyPackage.ts
@@ -2,7 +2,7 @@ import { InstallPackageData } from "../../types";
 import semver from "semver";
 
 /**
- * Prevent from installing legacy version (previous to remote signer support) for packages:
+ * From dappmanager v0.2.46 prevent from installing legacy version (previous to remote signer support) for packages:
  * - Prysm
  * - Prysm-Prater
  * - Gnosis-Beacon-Chain-Prysm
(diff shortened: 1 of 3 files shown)


3alpha (approved):

packages/dappmanager/src/modules/installer/ensureIsNotValidatorLegacyPackage.ts
```diff
@@ -0,0 +1,39 @@
...
+ * - Prysm
+ * - Prysm-Prater
+ * - Gnosis-Beacon-Chain-Prysm
+ */
+export async function ensureIsNotValidatorLegacyPackage(
+  packagesData: InstallPackageData[]
+): Promise<void> {
+  for (const pkg of packagesData) {
+    switch (pkg.dnpName) {
+      case "prysm-prater.dnp.dappnode.eth":
```
dapplion, line 15: Move this into a datastructure in src/params.ts

```ts
minimumAllowedPackageVersions: {
  "prysm-prater.dnp.dappnode.eth": "1.0.7"
}
```

pablomendezroyo: resolved

packages/dappmanager/src/modules/installer/ensureIsNotValidatorLegacyPackage.ts
```diff
@@ -11,29 +12,11 @@ export async function ensureIsNotValidatorLegacyPackage(
...
-      case "prysm.dnp.dappnode.eth":
-        if (semver.lte(pkg.semVersion, "1.0.0"))
-          throw Error(
-            `${pkg.dnpName}:${pkg.semVersion} is a legacy validator client, install a more recent version with remote signer support`
-          );
-        continue; */
-      default:
-        continue;
-    }
+    params.minimumAllowedPackageVersions.map(({ dnpName, version }) => {
```
dapplion, line 15: Array.map is used when you need to return values. Here you should use .forEach or for of operators

PR #988, opened 2022-03-18
ensure not old prysm installed when installing web3signer

pablomendezroyo:
Web3signer should not be allow to be installed if Prysm legacy is present


packages/dappmanager/src/modules/installer/ensureEth2MigrationRequirements.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Ensure the following requirements for Eth2 migration:
+ * - From dappmanager v0.2.46 prevent from installing legacy version (previous to remote signer support) for packages: (Prysm, Prysm-prater and Gnosis-Beacon-Chain-Prysm)
+ * -
+ */
+export async function ensureEth2MigrationRequirements(
+  packagesData: InstallPackageData[]
+): Promise<void> {
+  for (const pkg of packagesData) {
+    params.minimumAllowedPackageVersions.forEach(
```
dapplion, line 15: forEach cannot handle async functions, the thrown errors won't stop the installation

packages/dappmanager/src/modules/installer/ensureEth2MigrationRequirements.ts
```diff
@@ -0,0 +1,33 @@
...
+      async ({ dnpName, version }) => {
+        // Ensure is not a validator client legacy package
+        if (pkg.dnpName === dnpName && semver.lte(pkg.semVersion, version))
+          throw Error(
+            `${pkg.dnpName}:${pkg.semVersion} is a legacy validator client, install a more recent version with remote signer support`
+          );
+        // Ensure not installing web3signer if legacy client is installed
+        if (pkg.dnpName === params.web3SignerDnpName) {
```
dapplion, line 23: You don't need to put this inside the `params.minimumAllowedPackageVersions` loop

packages/dappmanager/src/modules/installer/ensureEth2MigrationRequirements.ts
```diff
@@ -0,0 +1,33 @@
...
+          const prysmPackage = await listPackageNoThrow({ dnpName: dnpName });
```
dapplion, line 24: Why is dnpName the prysm package?

packages/dappmanager/src/modules/installer/ensureEth2MigrationRequirements.ts
dapplion: Convert `params.minimumAllowedPackageVersions` into a `Map<string, string>` where key is dnpName, and value is minimum version. This way you don't have a quadratic loop

dapplion (changes requested):
See above

packages/dappmanager/src/params.ts
```diff
@@ -151,16 +151,21 @@ const params = {
...
     {
       dnpName: "prysm-prater.dnp.dappnode.eth",
       version: "0.1.7"
     }
   ],
 
-  // Web3signer dnpname
-  web3SignerDnpName: "web3signer-prater.dnp.dappnode.eth",
+  // Web3signer and clients dnpnames
+  web3SignerClientsDnpNames: [
```
dapplion, line 163: This is really a package exclusion list. A package excludes some other packages from being installable

pablomendezroyo: thats correct, a package within a version range exclude the installation of the web3signer for the network of the package

dapplion (approved):

PR #990, opened 2022-03-21
Changed Monero predefined exposable port

3alpha:
Fixing Monero predefined exposable port.

### DIFF
--- a/packages/dappmanager/src/modules/https-portal/exposable/predefined.ts
+++ b/packages/dappmanager/src/modules/https-portal/exposable/predefined.ts
@@ -146,7 +146,7 @@ export const exposablePredefined: {
       fromSubdomain: "monero",
       dnpName: "monero.dnp.dappnode.eth",
       serviceName: "monero.dnp.dappnode.eth",
-      port: 18082,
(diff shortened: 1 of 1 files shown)


alexpeterson91:
If you're changing the exposable ports here for monero why not merge my closed PR that actually made nethermind xDai exposable for both RPC and WS?  The manifest exposable HTTPS ports only work for one port in all attempts so far. Might as well just add it to the bootstrap list since you need to update the DAppManager again to fix the monero issue.

pablomendezroyo:
@alexpeterson91 @3alpha We must move the hardcoded exposable from the dappmanager to the packages manifest using https://docs.dappnode.io/es/developers/manifest-reference/#exposable

alexpeterson91:
> @alexpeterson91 @3alpha We must move the hardcoded exposable from the dappmanager to the packages manifest using https://docs.dappnode.io/es/developers/manifest-reference/#exposable

Lion said the same thing and chnsged the manifest. Several times. I never recall it working for both WS and RPC as it should. The manifest setup would only do the first listed exposed port

PR #997, opened 2022-03-28
Whitelist packages to use docker socket

pablomendezroyo:
## Context
Docker volumes are used by DAppNode packages in several different ways. There is a docker volume critical to be defined in the docker-compose, this is the docker socket volume: `/var/run/docker.sock`

## Solution
The docker socket volume is a critical volume that entails security vulnerabilities on the host. This docker socket volume should be only used by a well-known whitelist of packages:
- Dappmanager
- Core
- Vpn
- Wifi

## How to test it?
- Verify that installing the whitelist packages is allowed
- Verify that installing a package non-whitelisted with the docker socket volume result into an error

### DIFF
--- a/packages/dappmanager/src/modules/compose/unsafeCompose.ts
+++ b/packages/dappmanager/src/modules/compose/unsafeCompose.ts
@@ -66,6 +66,7 @@ const networkSafeKeys: (keyof ComposeNetwork)[] = [];
  * Strict sanitation of a docker-compose to prevent
  * - Use of uncontroled features
  * - Use of unsupported docker-compose syntax
+ * - Use of dokcer socket volume except for whitelisted packages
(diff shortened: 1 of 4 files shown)


packages/dappmanager/src/modules/compose/unsafeCompose.ts
```diff
@@ -202,7 +224,8 @@ function parseUnsafeVolumes(
 
   // External volumes are not allowed
   for (const [volName, vol] of Object.entries(volumes)) {
-    if (vol && vol.external)
+    if (!vol) continue;
+    if (vol.external)
```
dapplion, line 228: What's the reasoning of this change?

pablomendezroyo:
closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1078

PR #1001, opened 2022-03-31
Fix wrong check when installing web3signer

pablomendezroyo:
Fix a wrong check to not allow installing web3signer while having a legacy client installed in dappnode.


alexpeterson91:
Thank you!

PR #1005, opened 2022-04-04
Check for rest of the packages for HTTPS portal

3alpha:
## Context

Packages which have HTTPS mappings on update inject docker-compose which doesn't necessarily have external network specified

## Approach

During installation process, DNP_HTTPS is queried whether mapping exists, and if it does it adds external network to the compose. 

## Test instructions

- Do HTTPS mapping on a package
- Update that same package
- Check if mapping is broken after update
- Test specially when package === DNP_HTTPS and when it is not

### DIFF
--- a/packages/dappmanager/src/modules/https-portal/index.ts
+++ b/packages/dappmanager/src/modules/https-portal/index.ts
@@ -74,7 +74,7 @@ export class HttpsPortal {
     const httpsComposePath = ComposeEditor.getComposePath(params.HTTPS_PORTAL_DNPNAME, true)
     const editor = new ComposeEditor(ComposeEditor.readFrom(httpsComposePath))
 
-    if(editor.getComposeNetwork(externalNetworkName) === null) {
+    if (editor.getComposeNetwork(externalNetworkName) === null) {
       const httpsExternalAlias = getExternalNetworkAlias(httpsPortalContainer)
       addNetworkAliasCompose(httpsPortalContainer, externalNetworkName, [httpsExternalAlias])
     }
@@ -146,13 +146,13 @@ export class HttpsPortal {
     return mappings;
   }
 
-  async hasMapping(dnpName: string, serviceName: string) : Promise<boolean> {
+  async hasMapping(dnpName: string, serviceName: string): Promise<boolean> {
 
     const entries = await this.httpsPortalApiClient.list();
     const mappingAlias = getExternalNetworkAliasFromPackage(dnpName, serviceName);
     for (const { fromSubdomain, toHost } of entries) {
       const [alias, port] = toHost.split(":");
-      if(alias === mappingAlias) {
+      if (alias === mappingAlias) {
         return true;
       }
     }
--- a/packages/dappmanager/src/modules/https-portal/index.ts
+++ b/packages/dappmanager/src/modules/https-portal/index.ts
@@ -6,7 +6,7 @@ import {
 } from "../docker";
 import { listContainers } from "../docker/list";
 import params from "../../params";
-import { getExternalNetworkAlias } from "../../domains";
+import { getExternalNetworkAlias, getExternalNetworkAliasFromPackage } from "../../domains";
 import { PackageContainer, HttpsPortalMapping } from "../../types";
 import { HttpsPortalApiClient } from "./apiClient";
 import { addNetworkAliasCompose, removeNetworkAliasCompose } from "./utils";
@@ -144,6 +144,19 @@ export class HttpsPortal {
       }
     }
     return mappings;
+  }
+
+  async hasMapping(dnpName: string, serviceName: string) : Promise<boolean> {
+
+    const entries = await this.httpsPortalApiClient.list();
+    const mappingAlias = getExternalNetworkAliasFromPackage(dnpName, serviceName);
+    for (const { fromSubdomain, toHost } of entries) {
+      const [alias, port] = toHost.split(":");
+      if(alias === mappingAlias) {
+        return true;
+      }
+    }
+    return false;
   }
 
   private async getContainerForMapping(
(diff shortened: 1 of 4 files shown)


packages/dappmanager/src/modules/https-portal/index.ts
```diff
@@ -146,6 +146,19 @@ export class HttpsPortal {
     return mappings;
   }
 
+  async hasMapping(dnpName: string, serviceName: string) : Promise<boolean> {
+
+    const entries = await this.httpsPortalApiClient.list();
+    const mappingAlias = getExternalNetworkAliasFromPackage(dnpName, serviceName);
+    for (const { fromSubdomain, toHost } of entries) {
+      const [alias, port] = toHost.split(":");
+      if(alias === mappingAlias) {
```
dapplion, line 155: Can you run prettier on this code? Inconsistent format

packages/dappmanager/src/modules/installer/https.ts
```diff
@@ -0,0 +1,166 @@
...
+  pkg: InstallPackageData,
+  externalNetworkName: string
+): Promise<void> {
+  const networks = await dockerListNetworks();
+  if (!networks.find(network => network.Name === externalNetworkName)) {
+    await dockerCreateNetwork(externalNetworkName);
+  }
+
+  // Check whether DNP_HTTPS compose has external network persisted
+  const compose = new ComposeFileEditor(pkg.dnpName, true);
```
dapplion, line 25: Warning, this boolean means `isCore`, you must use `pkg.isCore`

packages/dappmanager/src/modules/installer/https.ts
```diff
@@ -0,0 +1,166 @@
...
+import { HttpsPortalMapping } from "../../common";
+import { getExternalNetworkAliasFromPackage } from "../../domains";
+import { ComposeFileEditor } from "../compose/editor";
+import { dockerListNetworks, dockerCreateNetwork } from "../docker";
+
+/**
+ * Recreate HTTPs portal mapping if installing or updating HTTPs package
+ */
+export async function httpsEnsureNetworkExists(
+  pkg: InstallPackageData,
```
dapplion, line 16: No need to pass `pkg` if you assert that it's the HTTPs below. Just use hardcoded values

packages/dappmanager/src/modules/installer/https.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+
+      const alias = getExternalNetworkAliasFromPackage(
+        pkg.dnpName,
+        serviceName
+      );
+      const aliases = [alias];
+      composeServiceEditor.addNetwork(externalNetworkName, { aliases });
+      compose.write();
+      break;
```
dapplion, line 72: Why is there a break here?

3alpha: We need to add external network once.

dapplion: I'm very confused by the code, could you explain what it's doing?

pablomendezroyo:
@3alpha what is the status of this PR?

packages/dappmanager/src/modules/installer/https.ts
3alpha: It adds external network to the HTTPS compose file on install/update if a mapping exists. This was cause of issues because mapping would exist, but on update HTTPS wouldn't reconnect to the external network.

pablomendezroyo:
We must consider defining the docker network `dnpublic_network` in the HTTPS compose file. This would simplify a lot everything, letting docker take care of the network ownership. See https://github.com/dappnode/DNP_HTTPS/issues/66

The network should only exist whenever the HTTPS package is installed and running. Defining the docker network `dnpublic_network` in the HTTPS compose file will ensure that the network is created together with the HTTPS container.

caveats:
- The `dnpublic_network` must be defined as internal in the  HTTPS compose
- There must be checked there are no issues with the dappnodeSDK and the dappmanager for this new internal network
- Portal mappings will be removed when the HTTPs package is removed, which makes sense
- To remove the HTTPS package could be necessary to manually disconnect every container connected to this network

packages/dappmanager/src/modules/installer/https.ts
```diff
@@ -106,32 +128,22 @@ export async function httpsExposeByDefaultPorts(
   }
 }
 
-async function hasRunningHTTPS(required: boolean = false) {
+// Utils
 
+/**
+ * Returns true if HTTPS package installed and running, otherwise return false
+ */
+async function isRunningHttps() {
```
3alpha, line 136: Would isHttpsRunning be better?

packages/dappmanager/src/modules/installer/https.ts
```diff
@@ -5,41 +5,55 @@ import params from "../../params";
...
-       if(!container.networks.find(n => n.name === externalNetworkName)) {
-        await dockerNetworkConnect(externalNetworkName, container.containerName, {Aliases: [alias]})
+  const containers =
+    (
+      await listPackageNoThrow({
+        dnpName: pkg.dnpName
+      })
+    )?.containers || [];
+
+  if (containers.length === 0) return;
```
3alpha, line 39: Is this necessary? As next line is pretty much the same condition.

packages/dappmanager/src/modules/installer/https.ts
pablomendezroyo: yes!

packages/dappmanager/src/modules/installer/https.ts
pablomendezroyo: I guess you are right but it is a nice way of working to always double check

pablomendezroyo (approved):

PR #1010, opened 2022-04-21
Limit resources feature

tropicar:
## Context

If some package has a problem because of the usage of resources, nethermind-xdai client, it would be interesting to be able to limit the amount of resources a package can use in order to protect the performance of the dappnode.
The feature is to let the dappmanager accepts the deploy option of docker-compose.

## Approach

This issue https://github.com/dappnode/DAppNode/issues/419

## Test instructions

In order to test this feature you need to install a package where you add these lines to the docker-compose:
```
    deploy:
      resources:
        limits:
          memory: 5000M
```
Then, once you have installed the package you need to access to the machine cli and type the command `docker stats`
and check the memory limit of the package you have instaled.

### DIFF
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -1,5 +1,7 @@
 // Aliases
 
+import { ResourceLimits } from "worker_threads";
+
 export type DnpName = string;
 
 // AUTH, SESSION types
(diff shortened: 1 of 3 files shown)


pablomendezroyo:
closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1078

PR #1012, opened 2022-04-23
*Implement driver support for the handshake chain

pcfreak30:
This adds a chain driver to support syncing status for the Handshake blockchain. Must be tested with the Handshake node package being developed.


packages/dappmanager/package.json
```diff
@@ -65,6 +65,7 @@
     "express": "^4.17.1",
     "express-fileupload": "^1.1.5",
     "helmet": "^4.4.1",
+    "hs-client": "^0.0.10",
```
dapplion, line 68: Since this library just does a simple GET request could you implement this driver without any additional NPM package? Long term we want to minimize external dependencies :pray:

pcfreak30: Not sure if it makes sense as it is the official chain library and it is making 2 different API calls. The code to do node-fetch is much more messy.

```
    const nodeInfo = (await fetch(`${apiUrl}:${webPort}`, {
            headers: {
                'Authorization': 'Basic ' + Buffer.from(username + ":" + password).toString('base64')
            }
        }
    )).json();
```

Thoughts?

dapplion: This looks simple enough to me!

packages/dappmanager/src/modules/chains/getChainDriverName.ts
```diff
@@ -18,5 +18,6 @@ const knownChains: { [dnpName: string]: ChainDriver } = {
   "bitcoin.dnp.dappnode.eth": "bitcoin",
   "monero.dnp.dappnode.eth": "monero",
   "prysm.dnp.dappnode.eth": "ethereum2-beacon-chain-prysm",
-  "prysm-prater.dnp.dappnode.eth": "ethereum2-beacon-chain-prysm"
+  "prysm-prater.dnp.dappnode.eth": "ethereum2-beacon-chain-prysm",
+  "handshake.dnp.dappnode.eth": "handshake"
```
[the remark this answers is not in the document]
dapplion: So this mapping of package name to driver was meant to boostrap the feature. If you include a driver property in the manifest of your package it will have the same effect.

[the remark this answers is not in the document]
pcfreak30: Done.

dapplion:
@pablomendezroyo could you review broken CI?

packages/dappmanager/src/modules/chains/getChainDriverName.ts
```diff
@@ -11,11 +11,24 @@ export function getChainDriverName(
...
+  "openethereum.dnp.dappnode.eth": "ethereum",
+  "ropsten.dnp.dappnode.eth": "ethereum",
+  "rinkeby.dnp.dappnode.eth": "ethereum",
+  "kovan.dnp.dappnode.eth": "ethereum",
+  "bitcoin.dnp.dappnode.eth": "bitcoin",
+  "monero.dnp.dappnode.eth": "monero",
+  "prysm.dnp.dappnode.eth": "ethereum-beacon-chain",
+  "prysm-prater.dnp.dappnode.eth": "ethereum-beacon-chain",
+  "lighthouse-prater.dnp.dappnode.eth": "ethereum-beacon-chain",
+  "teku-prater.dnp.dappnode.eth": "ethereum-beacon-chain"
```
dapplion, line 24: Please do not add packages here, see the comment below
```
  // ===============================
  // DO NOT ADD ANY NEW PACKAGE HERE
  // ===============================
  // Instead add "chain" property in your package, like https://github.com/dappnode/DAppNodePackage-prysm/pull/65
  // Full docs: https://docs.dappnode.io/developers/manifest-reference#chain
```

PR #1016, opened 2022-05-05
Only pin if is ipfs local

pablomendezroyo:
## Context

Ipfs was pinning content on every visit to the dappstore by the user. This is fine when using IPFS local but not when using IPFS remote.

## Approach

Only pin when using IPFS local.
## Test instructions

- Test that visiting the dappstore using IPFS local does pins the content
- Test that visiting the dappstore using IPFS remote does not pin the content


packages/dappmanager/src/modules/release/ipfs/downloadRelease.ts
```diff
@@ -73,7 +74,7 @@ async function downloadReleaseIpfsFn(
       );
 
       // Pin release on visit
-      ipfs.pinAddNoThrow(hash);
+      if (db.ipfsClientTarget.get() === "local") ipfs.pinAddNoThrow(hash);
```
dapplion, line 77: Doing a db call in the middle of IPFS module breaks separation in concerns a lot. Is there any nicer way to achieve the same?

PR #1018, opened 2022-05-05
Fix CI

pablomendezroyo:
## Context
- **Update https-ipfs-client dependency**
- **Update node engine version**: in Dockerfile and gha
- **Use of local ipfs node in int tests**
- **Get dappnode ipfs gateway id dynamically**
Int tests eventually need to startup a local IPFS node for testing, this local IPFS node needs to connect to the dappnode IPFS gateway. To accomplish that, it is required the node ID, which was been obtained statically which cause errors when this ID changes

## Approach

Get dappnode ipfs gateway ID dynamically

## Test instructions

- Ensure int tests works as expected with this change

### DIFF
--- a/packages/dappmanager/test/integrationSpecs/buildReleaseDirectory.ts
+++ b/packages/dappmanager/test/integrationSpecs/buildReleaseDirectory.ts
@@ -72,7 +72,8 @@ export async function uploadDirectoryRelease({
   if (disclaimer) writeAsset("disclaimer.md", disclaimer);
 
   const addResults = await ipfsAddAll(releaseDir);
-  const rootHash = addResults[addResults.length - 1].cid.toString();
+  // The last result is the root /test_files, the second is the dir /test_files/something/
+  const rootHash = addResults[addResults.length - 2].cid.toString();
 
   // Verify the uploaded files
   const files = await ipfs.list(rootHash);
--- a/packages/dappmanager/test/integrationSpecs/buildReleaseDirectory.ts
+++ b/packages/dappmanager/test/integrationSpecs/buildReleaseDirectory.ts
@@ -13,7 +13,7 @@ import {
   ComposeService
 } from "../../src/types";
 import { testDir, manifestFileName, composeFileName } from "../testUtils";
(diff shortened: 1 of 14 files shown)


packages/dappmanager/test/modules/ipfs/index.test.int.ts
```diff
@@ -31,11 +26,17 @@ describe("ipfs / integration test", function () {
   });
 
   it("Upload directory", async () => {
-    dirHash = await ipfsAddDirFromFs(dirPath);
+    const addResults = await ipfsAddAll(dirPath);
+    dirHash =
+      addResults
+        .find(addResult => addResult.path === "test_files/ipfs-test-upload")
+        ?.cid.toString() || "";
+    if (!dirHash) throw Error("No directory hash found");
```
dapplion, line 34: Abstract this into a separate function, which takes `dirPath` as argument and returns you the CID of the root directory

packages/dappmanager/test/integrationSpecs/buildReleaseDirectory.ts
```diff
@@ -71,7 +71,9 @@ export async function uploadDirectoryRelease({
   if (setupWizard) writeJson("setup-wizard.json", setupWizard);
   if (disclaimer) writeAsset("disclaimer.md", disclaimer);
 
-  const rootHash = await ipfsAddDirFromFs(releaseDir);
+  const addResults = await ipfsAddAll(releaseDir);
+  // The last result is the root /test_files, the second is the dir /test_files/something/
+  const rootHash = addResults[addResults.length - 2].cid.toString();
```
dapplion, line 76: Re-use function from https://github.com/dappnode/DNP_DAPPMANAGER/pull/1018/files#r868979902

packages/dappmanager/test/modules/ipfs/index.test.int.ts
pablomendezroyo: the returns from the `ipfsAdd` and the new `ipfsAddAll` have changed so the function `ipfsAddDirFromFs` did not work as expected

dapplion (approved):
LGTM! :fire:

Issue #1015, opened 2022-05-04
Change the way dappmanager installs docker-compose


tropicar:
Currently, the dappmanager Dockerfile installs docker-compose (https://github.com/dappnode/DNP_DAPPMANAGER/blob/master/Dockerfile) by downloading the binary from the dappnode's fork of the docker-compose repo (https://github.com/dappnode/compose). 

**Describe the solution you'd like**

- [x] Using the official repository of docker-compose (https://github.com/docker/compose/releases) instead of the dappnode's fork in order to download the docker-compose binary.

- [x] Check the major updates of the docker-compose that can affect the behaviour of the dappmanager.

- [ ] Update the docker-compose version in the dappmanager.

- [ ] Change how we install the docker-compose in the dappmanager. 

- [ ] Test the update of the docker-compose version of the dappamanager.

- [ ] Test the new installation process of the docker-compose version of the dappamanager.

**Additional context**

This installation process was originally so because dappnode requires compatibility with some architectures like aarch64. In the past, these architectures were not supported, but in the current versions, the official repository of docker-compose is compatible with this kind of architecture.

The differences in that time can be checked out  in the below links:
- Docker-compose official repository: https://github.com/docker/compose/releases/1.25.5
- Docker-compose dappnode fork: https://github.com/dappnode/compose/releases

tropicar:
Information about the big changes between docker-compose 1. x.x version (we use 1.25.x currently in our dappmanager) and 2. x.x(the latest is the 2.5.0).
- We are using as a version of the docker-compose file, the 3.5 file format, then the docker engine has to be  17.12.0+.
- Docker-compose v2 supports almost all the commands that docker-compose, the expectations are the next ones:

     1. `compose build --memory`
     2. `compose ps --filter KEY-VALUE`
     3. `compose rm --all`
     4. `compose scale`
     5. It will be possible to use `docker-compose cp` (and to copy files from a container to the host and vice versa)

We don't use any of these commands in the dappmanager, but we should test after the update how it works with the `docker-compose cp`. If this is possible we have to cap this feature for security reasons.

**Sources**:
- 2.0.0 release notes : https://github.com/docker/compose/releases/tag/v2.0.0
- Compose versioning: https://docs.docker.com/compose/compose-file/compose-versioning/
- Information about flags that will not be compatible: https://docs.docker.com/compose/cli-command-compatibility/
[...]

PR #1023, opened 2022-05-16
Update & install docker-compose in dappmanager

tropicar:
With these changes, the dappmanager installs docker-compose using the official repo of docker-compose and install the v2.5.0 version

This pr is part of this https://github.com/dappnode/DNP_DAPPMANAGER/issues/1015

> Provide context of this PR: why it was created? what does it fix? what new feature does it provide? ...

## Approach

This PR install version v2.5.0 in the dappmanager and change how the docker-compose is installed. Modify the dappmanager dockerfile.

## Test instructions

There is 2 kind of tests here:
- Test to check the dappmanager is using this version.
- Test to check a different kind of operations works with this new version of docker-compose.

## Check docker-compose version
Go into the dappnode and go inside of the dappmanager container and check the version or execute the next command:
`docker exec -it DAppNodeCore-dappmanager.dnp.dappnode.eth docker-compose -v`

## Check if the next operations work while you checks the logs of the dappmanager
- Install a package
- Restart a package
- Stop a package
- Start a package
- Remove package

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,13 +59,13 @@ RUN apk add --no-cache bind-tools docker
 #####################################
 FROM node:16.15.0-alpine
 
-ENV DOCKER_COMPOSE_VERSION 1.25.5
+ENV DOCKER_COMPOSE_VERSION v2.5.0
 
 RUN apk add --no-cache curl bind-dev xz libltdl miniupnpc zip unzip dbus bind \
   # See https://github.com/dappnode/DNP_DAPPMANAGER/issues/669
   avahi-tools
 
-RUN curl -L https://github.com/dappnode/compose/releases/download/$DOCKER_COMPOSE_VERSION/docker-compose-Linux-$(uname -m) > /usr/local/bin/docker-compose \
+RUN curl -L https://github.com/docker/compose/releases/download/$DOCKER_COMPOSE_VERSION/docker-compose-linux-$(uname -m) > /usr/local/bin/docker-compose \
   && chmod +x /usr/local/bin/docker-compose
 
 WORKDIR /usr/src/app


pablomendezroyo:
LGTM! 
@tropicar did you test that all the compose action performed by the dappmanager still works as expected? 

Also, do you think we should have the same compose version on the host and in the dappmanager container?

pablomendezroyo (approved):

tropicar:
> LGTM! @tropicar did you test that all the compose action performed by the dappmanager still works as expected?
> 
> Also, do you think we should have the same compose version on the host and in the dappmanager container?

I have tested the basic one I mentioned. I will create a list with all the different docker-compose operation that the dappmanager execute

pablomendezroyo:
> About the version in the host, the version in the host should be the same that the dappmanager.

Then please, open an issue in the installed referencing this

tropicar:
> > About the version in the host, the version in the host should be the same that the dappmanager.
> 
> Then please, open an issue in the installer referencing this
This issue is for update the docker-compose in the host https://github.com/dappnode/DAppNode/issues/431

dapplion (approved):

Dockerfile
```diff
@@ -59,13 +59,13 @@ RUN apk add --no-cache bind-tools docker
 #####################################
 FROM node:16.15.0-alpine
 
-ENV DOCKER_COMPOSE_VERSION 1.25.5
+ENV DOCKER_COMPOSE_VERSION v2.5.0
```
dapplion, line 62: Why does the version before didn't have the `v` prefix, but now it does?

tropicar: Because they have changed the nomenclature https://github.com/docker/compose/releases/tag/v2.5.1 and https://github.com/docker/compose/releases/tag/1.25.5

3alpha (approved):

PR #1026, opened 2022-05-21
Set max WPA characters for passphrase

### LINKED ISSUE #1024: It's possible to cause wifi to go into an endless loop if passphrase is too long
**Describe the bug**

WPA passphrases may be 8..63 characters long. Dappnode GUI allows to set a passphrase that is >63 characters. If this is done, the wifi module will not start but instead go into an endless loop until changed to something within the bounds of WPA.

**DAppNode version:**
Last version

alexpeterson91:
fixes #1024

## Context
Fixes #1024 (by making the max WPA passphrase character length an argument in the validator implemented the same as the minimum is, makes it so that this endless loop from #1024 to not be a possible situation by properly setting the max length, a warning appears when it exceeds 63 characters as the WPA standard describes, and it is baked into the validation for password strength function keeping it so that the warning not just pops up over 63 characters, but even when the password is properly copied into the second field, it shows a match, but the change password button remains inactive until the passphrase meets the strong password requirements, which now include the max character length in addition to the minimum.

## Approach

Add new variable for the max length of 63 characters and implemented it alongside  and within the password strength validation functions with front end GUI warnings and making it not possible to set a password that is too long for the WPA standard. and warning the user as soon as they exceed 63 characters

## Test instructions

Check Netlify and test.  IPFS hash will be uploaded by CI  for install too.

### DIFF
--- a/packages/admin-ui/src/utils/validation.ts
+++ b/packages/admin-ui/src/utils/validation.ts
@@ -42,7 +42,7 @@ export function validateStrongPassword(password: string): string | null {
     return `Password must be at least ${argMinLen} characters long`;
   }
    if (password.length > argMaxLen) {
-    return `Password must be less than ${argMaxLen} characters long`;
+    return `Password must not exeed ${argMaxLen} characters long`;
   }
   if (!/\d+/.test(password)) {
     return "Password must contain at least one number";
--- a/packages/admin-ui/src/utils/validation.ts
+++ b/packages/admin-ui/src/utils/validation.ts
@@ -1,4 +1,5 @@
 const argMinLen = 8;
+const argMaxLen = 63;
 
 export function validateDockerEnv(
   value: string,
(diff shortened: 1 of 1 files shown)


packages/admin-ui/src/utils/validation.ts
```diff
@@ -26,9 +27,22 @@ export function validateMinLength(
...
+  }
+  return null;
+}
+
 export function validateStrongPassword(password: string): string | null {
   if (password.length < argMinLen) {
     return `Password must be at least ${argMinLen} characters long`;
+  }
+   if (password.length > argMaxLen) {
+    return `Password must not exeed ${argMaxLen} characters long`;
```
dapplion, line 45:
```suggestion
    return `Password must not exceed ${argMaxLen} characters long`;
```

dapplion:
This character length limitation only applies to WPA passwords. Is the function validateStrongPassword used only for WPA passwords, or any password?

packages/admin-ui/src/utils/validation.ts
```diff
@@ -26,9 +27,22 @@ export function validateMinLength(
   return null;
 }
 
+export function validateMiaxLength(
```
dapplion, line 30: Is this function used somewhere?

```suggestion
export function validateMaxLength(
```

Issue #983, opened 2022-03-08
dappmanager should control who can use docker socket volume `/var/run/docker.sock:/var/run/docker.sock`


pablomendezroyo:
## Context 
The docker volume of the docker socket is a critical volume, that already a few packages such as WIFI and DAPPMANAGER have it (i.e https://github.com/dappnode/DNP_WIFI/blob/1fb93282b84ee23f5788645f90aff92c52e0d2d5/docker-compose.yml#L20)

## Issue
Allowing to share docker socket volume is a security vulnerability since docker is able to perform almost any action (it runs as sudo)

## Proposal
In order to avoid security vulnerabilities in DAppNode, we should consider controlling who can share this docker volume. Something like a whitelist of packages allowed to use this volume, if it is not whitelisted then throw an error when trying to install it

pablomendezroyo:
With SDK validation compose function, only a whitelist of dappnode packages can use the docker socket volume
closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1078

PR #1039, opened 2022-06-10
Forbid bind mounts

3alpha:
## Context

Bind mounts can be a security issue, especially when mounting important sockets like docker socket (see #983).

## Approach

 This PR would forbid all bind mounts since using bind mounts in dappnode packages have limited use.

## Test instructions

1) Go through all packages and see wheter bind mounts are used and research alternative if they are since this will break those packages
2) Try to install package with and without bind mount. Former should break, latter should go through.

For now, I will leave this as draft since after testing 1) we may have to add whitelist for some packages.

### DIFF
--- a/packages/dappmanager/src/modules/compose/unsafeCompose.ts
+++ b/packages/dappmanager/src/modules/compose/unsafeCompose.ts
@@ -201,10 +201,15 @@ function parseUnsafeVolumes(
 ): ComposeVolumes | undefined {
   if (!volumes) return undefined;
 
-  // External volumes are not allowed
+  // External volumes and bind mounts are not allowed
   for (const [volName, vol] of Object.entries(volumes)) {
-    if (vol && vol.external)
-      throw Error(`External volumes are not allowed '${volName}'`);
+    if (vol)
+    {
+      if(vol.external)
+        throw Error(`External volumes are not allowed '${volName}'`);
+      if (!vol.isNamed) // Volumes that are not named are bind mounts
+        throw Error(`Bind mounts are not allowed.`);
+    }
   }
 
   return mapValues(volumes, vol => pick(vol, volumeSafeKeys));
--- a/packages/dappmanager/src/modules/compose/unsafeCompose.ts
+++ b/packages/dappmanager/src/modules/compose/unsafeCompose.ts
@@ -23,6 +23,7 @@ import {
   Manifest
 } from "../../types";
 import semver from "semver";
+import { parseVolumeMappings } from "./volumes";
 
 interface ValidationAlert {
   name: string;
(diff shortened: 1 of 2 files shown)


pablomendezroyo:
Also, I am a bit confused about the function https://github.com/dappnode/DNP_DAPPMANAGER/blob/17fc51007703e4c04fa4bc3d92a529dca6dafa49/packages/dappmanager/src/modules/compose/volumes.ts#L23 in order to detect if there are or not docker volumes.

For me, the simplest way of detecting docker bind-mounted volumes is: every docker volume defined at the compose service level MUST be also defined at the top level, otherwise it is a bind-mounted. For example:
* Not bind-mounted
```
version: "3.5
services:
  example:
    ....
    volumes:
      - test:/some/path
volumes:
  - test: {}
```
* Bind mounted
``` 
version: "3.5
services:
  example:
    ....
    volumes:
      - test:/some/path
```

3alpha:
>For me, the simplest way of detecting docker bind-mounted volumes is: every docker volume defined at the compose service level MUST be also defined at the top level, otherwise it is a bind-mounted. For example:


It's not like that, your example won't work if you try to run it for bind mounts. Bind mounts on lefthand side of `:` must have path, that is how they are parsed. (I just noticed an issue in `parseVolumeMappings` function which I will fix).


CI needs to be fixed because on some core packages we need bind mounts (DAppManager need mount to docker socket for example). So I am wondering what is the best way to do this. Do we specify whitelist or can this validation be done elsewhere where it goes around core packages?

pablomendezroyo:
> It's not like that, your example won't work if you try to run it for bind mounts. Bind mounts on lefthand side of : must have a path, that is how they are parsed

You could name a volume name with `/` so this function would be unuseful? the point is that every service volume must be defined at the top level of the docker compose volumes section. This will ensure that it will not be bind-mounted

pablomendezroyo:
> It's not like that, your example won't work if you try to run it for bind mounts. Bind mounts on lefthand side of : must have a path, that is how they are parsed

You could name a volume name with `/` so this function would be unuseful? the point is that every service volume must be defined at the top level of the docker compose volumes section. This will ensure that it will not be bind-mounted

3alpha:
For example you can try to run following composes
```
version: "3.5"
services:
  example:
    image: busybox
    volumes:
      - /tmp:/tmp
volumes:
  /tmp: {}
```
And
```
version: "3.5"
services:
  example:
    image: busybox
    volumes:
      - test:/tmp
volumes:
  test: {}
```
And
```
version: "3.5"
services:
  example:
    image: busybox
    volumes:
      - nottest:/tmp
volumes:
  test: {}
```

Only second one is valid.

PR #1045, opened 2022-06-15
Fix Dashboard ChainDriver to recognize non-standard ports defined in package manifests

Nabsku:
## Context

If you're running an Ethereum (EL) node on a non-standard port, you'll not get an integration in the dashboard - even if you specify the chainDriver in the package manifest like this:
 
```
 "chain": {
    "driver": "ethereum",
    "serviceName": "arbitrum-node",
    "portNumber": 8547
  },
```

## Approach

This PR fixes this by importing the ChainDriverSpecs and adding a port variable.

## Test instructions

Deploy EL client with non-standard port mapping, with chainDriver specified in package manifest.


pablomendezroyo (dismissed):

packages/dappmanager/src/modules/chains/drivers/ethereum.ts
```diff
@@ -23,14 +24,16 @@ const gethSyncHelpUrl = whyDoesGethTakesSoMuchToSync;
...
+  chainDriver: ChainDriverSpecs
 ): Promise<ChainDataResult> {
   const container = dnp.containers[0];
   if (!container) throw Error("no container");
   const containerDomain = getPrivateNetworkAlias(container);
 
+  const port = chainDriver.portNumber || 8545;
   // http://ropsten.dappnode:8545
-  const apiUrl = `http://${containerDomain}:8545`;
+  const apiUrl = `http://${containerDomain}:${port}`;
```
pablomendezroyo, line 36: Here you are using the `containerDomain` obtained from the first container which is not ideal. Instead, you should try to use first the `serviceName` from the `chainDriver`, and if is not defined, then use the other approach. Take as reference https://github.com/dappnode/DNP_DAPPMANAGER/blob/8eaaff5ddf0ffacac58ba9ef1f4cb5ad97d26676/packages/dappmanager/src/modules/chains/drivers/ethereum2.ts#L17

Nabsku: Looking into it, thanks for the review

Nabsku: Done, tested it with and without chain declaration in package manifests.

PR #1049, opened 2022-06-17
Import types from `@dappnode/dappnodesdk`

pablomendezroyo:
## Context

There are multiple types/interfaces duplicated between dappmanager and dappnodesdk, which is not ideal. Specially from the files `setup-wizard` `manifest` and `compose`

## Approach

Use the `@dappnode/dappnodesdk` as a dependency to import the types and remove the types declared in the dappmanager

## Test instructions

Make sure nothing is broken related to the types/interfaces mentioned


packages/admin-ui/src/common/types.ts
```diff
@@ -1,3 +1,6 @@
+import { Compose } from "@dappnode/dappnodesdk/src/files/compose";
+import { ChainDriver } from "@dappnode/dappnodesdk/src/files/manifest";
```
dapplion, line 2: You don't want to import deep from inside a module, this things should be exported from the root file

Issue #1055, opened 2022-06-20
Migration from openethereum users to another client


tropicar:
**Is your feature request related to a problem? Please describe.**

The file migrateEthchain.ts use an old feature to instal the client. In this file appears openethereum references and this function is executed by the dappmanager in the index. 
We should discuss how we are dealing with the users currently using openEthereum client.

**Describe the solution you'd like**
We should discuss this topic in devs meeting. I dont have an idea to solve this. But we should talk about :
1. Will we change the user client from openethereum to another with no ask? this does not seem the best solution.
2. Can we ask the users when they update the dappmanager?

**Additional context**

OpenEthereum Client will be deprecated when the merge happens, even this client will not be prepared for the merge, so its very important to retire this client option as soon as possible.

pablomendezroyo:
closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1056 and https://github.com/dappnode/DNP_DAPPMANAGER/pull/1059

PR #1056, opened 2022-06-20
Remove openethereum client

tropicar:
## Context

Openethereum will be deprecated, it will not be available in the dappstore and users must start migrating to other clients

## Approach

Startup migration to deprecate openethereum. It will remove the package with all the volumes and if the repository selected was openethereum, change the `ethClientTarget` based on the following preference: geth/light-geth > nethermind > remote
Remove all migration of ethchain. Ethchain is not longer in the dappstore and last publishe was 2 years ago

## Test instructions

- Install open Ethereum and select the repository client openethereum. Then update to a dappmanager with this code. Make sure the open Ethereum was removed with the volumes and the client selected in the repository was set to remote
- Install openEthereum and  geth full node. Then update to a dappmanager with this code. Make sure the open Ethereum was removed with the volumes and the client selected in the repository was set to geth full node
- Install openEthereum and  geth light client. Then update to a dappmanager with this code. Make sure the open Ethereum was removed with the volumes and the client selected in the repository was set to geth light client
- Install openEthereum and  nethermind. Then update to a dappmanager with this code. Make sure the open Ethereum was removed with the volumes and the client selected in the repository was set to nethermind


tropicar:
I have created this issue(https://github.com/dappnode/DNP_DAPPMANAGER/issues/1055) because removing or touching the migrateEthchain file can provoke breaking changes and we should discuss what to do with the users who are using openthereum client.

tropicar (commented):
LGTM
In the test, if you have installed OpenEthereum as repository, it changes to remote after the update, I  think maybe we should show some kind of notification.

pablomendezroyo (approved):

PR #1061, opened 2022-06-27
Allow installing web3signer if it is a Prysm legacy update

pablomendezroyo:
## Context

Installing web3signer should be allowed to be installed when it is installed as a dependency and the Prysm Legacy is been updated to a non-legacy version

## Approach

Do not throw error when this situation matches

## Test instructions

Having Prysm legacy installed, try to update it to a non-legacy version with web3signer as a dependency. It should be allowed


tropicar (commented):
LGTM

PR #1076, opened 2022-07-14
Add besu client option

Nabsku:
## Context

Enables the user to select the Ethereum EL client "Besu" as a fullnode client in the repository overview.

## Approach

I developed a Besu package, tested that for a good while and then implemented the change in this PR. Besu responds to fullnode.dappnode and gets used as repo source now.

## Test instructions

Choose "Besu" as EL full node in the repository settings, install besu and wait for it to sync.

### DIFF
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -1311,13 +1311,14 @@ export enum IpfsClientTarget {
  * Eth provider / client types
  * Manage the Ethereum multi-client setup
  */
-export type EthClientTargetPackage = "geth-light" | "geth" | "nethermind";
+export type EthClientTargetPackage = "geth-light" | "geth" | "nethermind" | "besu";
 export type EthClientTarget = EthClientTargetPackage | "remote";
 export const ethClientTargets: EthClientTarget[] = [
   "remote",
   "geth-light",
   "geth",
-  "nethermind"
+  "nethermind",
+  "besu"
 ];
 
 /**
--- a/packages/admin-ui/src/components/EthMultiClient.tsx
+++ b/packages/admin-ui/src/components/EthMultiClient.tsx
@@ -29,6 +29,8 @@ export function getEthClientPrettyName(target: EthClientTarget): string {
       return "Geth";
     case "nethermind":
       return "Nethermind";
+    case "besu":
+      return "Besu";
   }
 }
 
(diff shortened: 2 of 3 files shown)


pablomendezroyo:
@tropicar @eduadiez Do we want to include the client Besu into the repository eth options in dappnode?

alexpeterson91:
> @tropicar @eduadiez Do we want to include the client Besu into the repository eth options in dappnode?

I say yes, I don't really know much about the client but we've seen some major implementations fold, go EoL, or just have major resource issues, the more the better I say but i dont know all the costs and benefits adding it would do.  So I'm with @Nabsku and say we add it, thanks for building it man.  But as with everything we should probably put it through some more testing i can spin it up on a couple of my machines to see how it works.

Nabsku:
My rationale behind this was that we should strive to provide as many options as possible, since having a supermajority on the EL side is as bad if not worse than a supermajority on the CL side.

Besus new bonsai storage trie reduces the amount of storage needed for a full node and doesn't need pruning (major UX improvement!). Feel free to ping me regarding the besu package, I'm happy to provide information and improve the package as needed!

tropicar:
I agree that is a good option to add it but we have to talk about:
- this package will be public? or official?
- Will be maintained by us? I mean dappnode publishes the new releases etc
I agree that more options = better for all but we have to talk about this

alexpeterson91:
My opinion is start with it being public and maintained and championed by @Nabsku as he's already indicated he's open to doing. Public but still in the DAppStore. And see how users react. I think it could be a great option for some users. If it gains a lot of traction then we discuss folding it into an official DNP. 
Again all after deep testing. 

That's my 2 cents. Anyways I'm gonna start testing the client myself on a couple machines today.

Nabsku:
@alexpeterson91 we now have duplicate PRs regarding this. #1115 and this one here.

alexpeterson91:
> @alexpeterson91 we now have duplicate PRs regarding this. #1115 and this one here.

Noted: Labeled my draft PR as `duplicate`

alexpeterson91 (commented):
Just need to publish the package as a DNP @eduadiez i don't know who should be publishing it and how to add it to the store. Once it's tested it'll be added to the store and as a permanent option in the enum which is lasting only as long as the stakers ui remains unpublished for each network.

pablomendezroyo:
Closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1158

PR #1081, opened 2022-07-26
Add check for volume sizes

3alpha:
## Context

Connected with this: https://github.com/dappnode/DAppNode/issues/417

## Approach

Since showing volume size for bind volumes is not feasible since it has to be calculated, we don't show that. However, total volume size is shown, but counting as 0 volume size of bind mount volumes. This PR removes total volume size in those cases since that is obviously wrong information.

## Test instructions

In the volumes section of package check whether:
1) total volume size is shown for packages that don't have bind mounts (custom location for volumes)
2) total volume size is not show for packages that have bind mounts

### DIFF
--- a/packages/admin-ui/src/pages/packages/components/Info/VolumesList.tsx
+++ b/packages/admin-ui/src/pages/packages/components/Info/VolumesList.tsx
@@ -79,6 +79,8 @@ export const VolumesList = ({ dnp }: { dnp: InstalledPackageDetailData }) => {
     return null;
   }
 
+  const isTotalVolumeSizeIllDefined = volumes.map(v => v.size).includes(undefined);
+
   const totalVolumeSize = volumes.reduce(
     (total, vol) => total + (vol.size || 0),
     0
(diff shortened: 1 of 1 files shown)


packages/admin-ui/src/pages/packages/components/Info/VolumesList.tsx
```diff
@@ -79,6 +79,8 @@ export const VolumesList = ({ dnp }: { dnp: InstalledPackageDetailData }) => {
     return null;
   }
 
+  const isTotalVolumeSizeIllDefined = volumes.map(v => v.size).includes(undefined);
```
pablomendezroyo, line 82: Is the word `isTotalVolumeSizeIllDefined` a typo? what does it mean ...size ill defined ?

3alpha: Ill defined means not well defined. But we can change that in something more readable.

pablomendezroyo (approved):

pablomendezroyo (approved):

PR #1099, opened 2022-08-04
Set `globalEnvs` (as `string [ ]` and `all: true`) in the `docker-compose.yml`

pablomendezroyo:
## Context

From https://github.com/dappnode/DAppNodeSDK/pull/266, the `globalEnvs` manifest feature allows defining which global envs add to which service. 

## Approach

Implement the addition of the global envs to the compose file based on this feature. There can be added two types of global envs:
- **all**: this is the current feature implemented, it injects the `.env` file into the `docker-compose`
- Defining envs and services to set those vars

## Test instructions

Install two packages:
1. Package with a manifest defining the `globalEnvs` like:
```
globalEnvs?: {
        all?: boolean;
    } 
``` 
2. Package with a manifest defining the `globalEnvs` like:
```
globalEnvs?: {
        envs: string[];
        services: string[];
    }[ ]
```

Make sure that in both cases the `docker-compose.yml` of the given package was edited and added the envs (env file for the 1 and env one by one to the 2)

### DIFF
--- a/packages/admin-ui/package.json
+++ b/packages/admin-ui/package.json
@@ -21,7 +21,7 @@
     "mock-standalone:build": "REACT_APP_MOCK=true yarn build"
   },
   "dependencies": {
-    "@dappnode/dappnodesdk": "^0.2.64",
+    "@dappnode/dappnodesdk": "^0.2.65",
     "@reduxjs/toolkit": "^1.3.5",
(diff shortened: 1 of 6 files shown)


pablomendezroyo:
:white_check_mark: Tested

PR #1108, opened 2022-08-18
Set HTTPS mappings to local dns

pablomendezroyo:
## Context

Set HTTPS portal mappings to the local dns. This will allow to use HTTPS portal mappings without having to expose it to the internet

## Test instructions

Expose a HTTPS portal mapping and make sure that it is available from inside and outside with `dig` command

### DIFF
--- a/packages/dappmanager/src/modules/https-portal/index.ts
+++ b/packages/dappmanager/src/modules/https-portal/index.ts
@@ -126,7 +126,10 @@ export class HttpsPortal {
     removeNetworkAliasCompose(container, externalNetworkName);
 
     // Removes from the bind the domain:https-IP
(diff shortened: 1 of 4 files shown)


3alpha (commented):
Why not just adding `*.{dyndnsIdentity}` to BIND and thats it?

pablomendezroyo:
> Why not just adding `*.{dyndnsIdentity}` to BIND and thats it?

But dont we want to just add the https mapping of the package for which the mapping has been created?

3alpha:
We need to have DNS name for mapping that exist, but if we have DNS names for mappings that do not exist, that is not a problem. For example, even right now every domain `*.1234abc.dyndns.dappnode.io` from outside your DAppNode will resolve to your DAppNode's public IP, but that is not an issue.

PR #1115, opened 2022-08-27
Add Besu as official EL and as Main Repo Client

alexpeterson91:
added Besu

## Context

With the upcoming merge around the corner, we finally got multi-client implemented on all chains, solving the consensus layer diversification issue.  This aims to solve the new execution layer diversification issue created by the following:
- Deprecation of OpenEthereum
- Instability of Erigon
- Recent major bug in Geth, the majority client which made Nethermind the only option for a short time.
- Nethermind's resource hogging issue is still in the code AFAIK, but is being effectively mitigated by the ulimits now imposed by Docker for nethermind containers; this solution is not sufficient for very low RAM machines, Besu is touted to and has shown in testing to be more resource friendly than Nethermind.

## Approach

Will be adding Besu as an option alongside Nethermind and Geth as global ENVs for EL and CL client matching.
Also adding Besu as an option to be an official Repo client.

## Test instructions

Testing instructions on Notion.

### DIFF
--- a/packages/admin-ui/src/common/types.ts
+++ b/packages/admin-ui/src/common/types.ts
@@ -1000,13 +1000,14 @@ export enum IpfsClientTarget {
  * Eth provider / client types
  * Manage the Ethereum multi-client setup
  */
-export type EthClientTargetPackage = "geth-light" | "geth" | "nethermind";
+export type EthClientTargetPackage = "geth-light" | "geth" | "nethermind" | "besu";
 export type EthClientTarget = EthClientTargetPackage | "remote";
 export const ethClientTargets: EthClientTarget[] = [
   "remote",
   "geth-light",
   "geth",
-  "nethermind"
+  "nethermind",
+  "besu"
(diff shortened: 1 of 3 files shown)


pablomendezroyo:
Great PR description!

Is Besu package published tested and under dappnode org control?

alexpeterson91:
> Great PR description!
> 
> 
> 
> Is Besu package published tested and under dappnode org control? 

Thanks idk why I felt like actually writing at the time. Wrote great comments in my code that day too. Idk if I have written one since. Lol. And well I think it needs to be published by Edu but it's pinned withthe dappnode piñata account and yes it's in the organization. 
The package is all good and should be added to the DAppStore I just wanna test the ARM64 build to confirm it works and doesn't just build that's the only open PR on it right now.

alexpeterson91:
It needs to go through the full testing regimen like all other clients. I don't know where thats being tracked right now.

PR #1131, opened 2022-09-15
Fix missing `await` in promise

pablomendezroyo:
Fix missing `await` in promises

### DIFF
--- a/packages/dappmanager/src/modules/release/index.ts
+++ b/packages/dappmanager/src/modules/release/index.ts
@@ -5,14 +5,20 @@ import { PackageRelease, PackageRequest } from "../../types";
 import dappGet, { DappgetOptions } from "../dappGet";
 import { Apm } from "../apm";
 import { Manifest } from "@dappnode/dappnodesdk";
+import { DappGetState } from "../dappGet/types";
 
 export class ReleaseFetcher extends Apm {
   /**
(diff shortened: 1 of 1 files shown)


packages/dappmanager/src/modules/release/index.ts
```diff
@@ -5,14 +5,20 @@ import { PackageRelease, PackageRequest } from "../../types";
...
-  async resolveReleaseName(name: string, version?: string) {
-    return resolveReleaseName(this, name, version);
+  async resolveReleaseName(
+    name: string,
+    version?: string
+  ): Promise<{
+    hash: string;
+    origin?: string | undefined;
+  }> {
+    return await resolveReleaseName(this, name, version);
```
dapplion, line 21: This await is not necessary

dapplion:
You should add this eslint rule to ensure you are covering all cases https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md

packages/dappmanager/src/modules/release/index.ts
pablomendezroyo: Right, the issue comes when adding the eslint rule, so I return a promise in a function that is not typed to return a promise. So when I use such function the linter does not warn you to use `await`

PR #1138, opened 2022-09-23
add besu goerli to stakersUI

alexpeterson91:
Adding Besu-Goerli to Stakers


packages/dappmanager/src/modules/stakerConfig/utils.ts
```diff
@@ -95,7 +95,8 @@ export function getNetworkStakerPkgs(network: Network): {
           {
             dnpName: "goerli-nethermind.dnp.dappnode.eth",
             minVersion: "1.0.1"
-          }
+          },
+          { dnpName: "besu-goerli.dnp.dappnode.eth", minVersion: "0.1.0" }
```
pablomendezroyo, line 99: Could you change the dnp name and repo to be consistent with the naming of the other goerli exec clients (goerli-geth and goerli-erigon) from `besu-goerli` to `goerli-besu`

pablomendezroyo (changes requested):

packages/dappmanager/src/modules/stakerConfig/utils.ts
alexpeterson91: Yes will do.

alexpeterson91: @Nabsku can yiu do this to the package and the 5 CLs? I'll fix here. And we'll pick up in a few more hours

alexpeterson91:
goerli Besu must be published before merging,

Nabsku:
changed all CLs to `goerli-besu` instead of `besu-goerli`!

alexpeterson91 (commented):
I'm gonna put together hashes of the packages needed for this all to work properly there's a pr for each CL and the W3S. And since adding clients to the stakersUI now requires all of this lodestar prater is good to go. Lion signed off on it. So we may want to add them both to avoid having to do multiple core and all client updates. Anyways I'll have hashes to test both of the clients posted this week.

3alpha (approved):

pablomendezroyo:
Whats the current status of this?

pablomendezroyo (approved):

Issue #1130, opened 2022-09-15
Set dynamically the memory limits deppending on the host machine resources


pablomendezroyo:
Instead of having the host resources limits hardcoded, determine it deppending on the host machine resources. Determine and pick a proper percentage.

See https://github.com/dappnode/DNP_DAPPMANAGER/blob/0af26e4872c5cefa442a44109867a4072a398045/packages/dappmanager/src/modules/compose/setDappnodeComposeDefaults.ts#L40

PR #1140, opened 2022-09-28
Delete default limit

3alpha:
Default memory limit has been giving us more trouble then it solved issues so we should remove it.

### DIFF
--- a/packages/dappmanager/src/modules/compose/setDappnodeComposeDefaults.ts
+++ b/packages/dappmanager/src/modules/compose/setDappnodeComposeDefaults.ts
@@ -34,13 +34,6 @@ export function setDappnodeComposeDefaults(
       (serviceUnsafe, serviceName) => {
         return sortServiceKeys({
           // OVERRIDABLE VALUES: values that in case of not been set, it will take the following values
-          deploy: {
-            resources: {
-              limits: {
-                memory: "16384M"
-              }
-            }
-          },
           logging: {
             driver: "json-file",
             options: {
--- a/packages/dappmanager/test/modules/compose/setDappnodeComposeDefaults.test.ts
+++ b/packages/dappmanager/test/modules/compose/setDappnodeComposeDefaults.test.ts
@@ -108,7 +108,6 @@ describe("setDappnodeComposeDefaults", () => {
         "beacon-chain": {
           container_name:
             "DAppNodePackage-beacon-chain.teku-gnosis.dnp.dappnode.eth",
-          deploy: { resources: { limits: { memory: "16384M" } } },
           dns: "172.33.1.2",
           environment: {
             LOG_TYPE: "INFO",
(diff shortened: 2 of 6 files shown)


alexpeterson91:
I agree wholeheartedly this is definitely the source of geth issues and all my Java issues on my servers with 128gb ram and 36 cores this is very important. I think we need to do something else but this global limit as it's currently written is causing an enormous amount of support cases and upset users that just can't sync their stuff anymore and are losing money by the epoch. We need to remove it now and come up with a dynamic limit or just apply limits to packages we have identified unfixed memory leaks in.

pablomendezroyo:
We could define the default resource limit dynamically, check this out https://github.com/dappnode/DNP_DAPPMANAGER/issues/1130

For critical packages we can still define a higher limit at the docker compose, and that value will not be overwritten. This will fix the issue

3alpha:
This default limit has solved zero problems and caused lots of issues. Some packages become unstable because they get over the limit for a very short amount of time and get OOM killed even thought they behave otherwise fine.

If there is a package that has memory leak and it causes issues, it is a 3 minute job to release it with memory limit.

alexpeterson91:
> @3alpha @alexpeterson91
> 
> We could define the default resource limit dynamically, check this out #1130
> 
> For critical packages we can still define a higher limit at the docker compose, and that value will not be overwritten. This would fix the issue.

I'm with Luka on this one, while i did initially agree with your proposal of dynamically allocating resources, I've spent many weeks testing all my machines and doing an incredible amount of support, and I haven't had any issues with RAM even syncing multiple mainnet clients simultaneously  from a fresh install with nothing more than removing the global static limit.

- I thought a percentage would work well initially because it would help users with low RAM and high RAM the same. 
- But what I have noticed in practice is that when syncing things this would cause exactly the same issues.  
- I thought like 25% would work well but just watching Grafana during syncing and other installs for days, and I found that certain packages really eat up RAM during initial sync (especially depending on the syncmode) but don't need too much RAM once it's synced.  
- Also some packages lean heavily on one resource like Avalanche syncing uses very minimal memory, but a lot of CPU, others lean on memory hard.
- Besu was a great example during initial sync on a 64GB NUC, used 35-40% of RAM during sync in DB high spec mode and X_Snap for sync-mode which takes a lot of RAM.  But once it was synced that number dropped to ~10%. 
- It's also still going to cause major issues for users with 16GB or less (32GB needs to be the minimum spec and 64GB the recommended) 25% of 16GB is just 4GB which isn't enough for many packages to sync at all.  

There are packages with memory leaks and other things that we can tweak in the package easily without having to do anything in the DAPPMANAGER which is so much more invasive, dangerous and takes much more time, I agree 100% with Luka and we should remove all global limits on RAM and only implement them package by package when a leak is identified as we did with Nethermind.

Separately @pablomendezroyo I just noticed in the last week that you hardcoded JVM max heap file limits in all Teku clients I dont generally have issues with Teku as a CL on any chain (haven't tried with hundred or so Gnosis validators yet but it's quite stable as is. But i also don't think that it should have a hardcoded max heap file size of 4GB, I think that's reasonable and should be left as default, but a new `JVM_OPTS` ENV to set a different min or max heap size or threads/etc or other JVM settings should be added to the advanced config editor prefilled with the `--Xmx4G` currently in all teku clients. Besu has it in the setup wizard as BESU_OPTS but i think there too it should be moved to just the advanced config editor.

pablomendezroyo (approved):

PR #1173, opened 2022-10-30
Prune`userActionLogs`

pablomendezroyo:
- Truncate the file `userActionLogs.json` to 0 if the size is greater than 4MB. Files with size higher are dangerous to parse and may result into unexpected behaviours.
- Do not allow to write new logs to the file if the size of the log exceeds 3KB.


3alpha:
Why delete everything and not just old ones?

pablomendezroyo:
> Why delete everything and not just old ones?

Good point! I was thinking about to do the same, however Ive seen this file size may be up to 150 MB. Parsing this file tryng to keep new ones may result into a long process an even few tools like `JSON.parse / JSON.stringiy` may crash due to the length of the data

PR #1180, opened 2022-11-07
Cache staker ui packages metadata

pablomendezroyo:
Cache the stakers ui packages metadata to speed up the initial loading. Update the cache on every request in background

### DIFF
--- a/packages/dappmanager/src/calls/packageRemove.ts
+++ b/packages/dappmanager/src/calls/packageRemove.ts
@@ -87,7 +87,15 @@ export async function packageRemove({
     const containerNames = dnp.containers.map(c => c.containerName);
     await Promise.all(
       containerNames.map(async containerName => {
-        await dockerContainerStop(containerName, { timeout });
+        // Continue removing package even if container is already stopped
+        await dockerContainerStop(containerName, { timeout }).catch(e => {
+          if (
+            e.reason.includes("container already stopped") &&
+            e.statusCode === 304
+          )
+            return;
+          else throw e;
+        });
         await dockerContainerRemove(containerName, { volumes: deleteVolumes });
       })
     );
--- a/packages/dappmanager/src/db/stakerConfig.ts
+++ b/packages/dappmanager/src/db/stakerConfig.ts
@@ -4,16 +4,28 @@ import {
   ConsensusClientGnosis,
   ExecutionClientGnosis,
   ExecutionClientMainnet,
-  ExecutionClientPrater
+  ExecutionClientPrater,
+  StakerItemMetadata
 } from "../types";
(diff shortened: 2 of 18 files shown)


packages/dappmanager/src/db/stakerConfig.ts
```diff
@@ -4,11 +4,23 @@ import {
...
 
+// Cache
+
+const STAKER_ITEM_METADATA = "staker-item-metadata";
+
+export const stakerItemMetadata = dbCache.indexedByKey<StakerItemData, string>({
+  rootKey: STAKER_ITEM_METADATA,
+  getKey: target => target,
+  validate: (id, metadata) =>
+    typeof id === "string" && typeof metadata === "object"
```
dsimog01, line 21: Maybe you can specify more the type of metadata?

packages/dappmanager/src/db/stakerConfig.ts
```diff
@@ -4,11 +4,23 @@ import {
...
-import { dbMain } from "./dbFactory";
+import { dbCache, dbMain } from "./dbFactory";
 import { interceptGlobalEnvOnSet } from "./interceptGlobalEnvOnSet";
 
+// Cache
+
+const STAKER_ITEM_METADATA = "staker-item-metadata";
+
+export const stakerItemMetadata = dbCache.indexedByKey<StakerItemData, string>({
+  rootKey: STAKER_ITEM_METADATA,
```
dsimog01, line 18: Why don't you just put "staker-item-metadata" here? It seems you don't use the constant in any other place.

pablomendezroyo: Checkout the rest of db files, they use that standard. So in order to not break the standard I keep it like that

packages/dappmanager/src/db/stakerConfig.ts
pablomendezroyo: the js function `typeof` is meant for native js types: `string` `boolean` `number`, etc.

packages/dappmanager/src/calls/packageRemove.ts
```diff
@@ -87,7 +87,15 @@ export async function packageRemove({
     const containerNames = dnp.containers.map(c => c.containerName);
     await Promise.all(
       containerNames.map(async containerName => {
-        await dockerContainerStop(containerName, { timeout });
+        // Continue removing package even if container is already stopped
+        await dockerContainerStop(containerName, { timeout }).catch(e => {
+          if (
+            e.reason.includes("container already stopped") &&
```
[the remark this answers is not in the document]
pablomendezroyo: Agree!

PR #1187, opened 2022-11-09
Set Besu as `public` instead as `dnp`

pablomendezroyo:
Keep the Besu package as `public`. https://github.com/dappnode/DAppNodePackage-besu

### DIFF
--- a/packages/admin-ui/src/__mock-backend__/stakerConfig.ts
+++ b/packages/admin-ui/src/__mock-backend__/stakerConfig.ts
@@ -42,14 +42,14 @@ export const stakerConfig: Pick<
             },
             {
               status: "ok",
-              dnpName: "besu.dnp.dappnode.eth",
+              dnpName: "besu.public.dappnode.eth",
               isInstalled: true,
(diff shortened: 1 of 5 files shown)


packages/admin-ui/src/__mock-backend__/stakerConfig.ts
```diff
@@ -212,7 +212,7 @@ export const stakerConfig: Pick<
               isSelected: true,
               avatarUrl: "",
               metadata: {
-                name: "besu.dnp.dappnode.eth",
+                name: "goerli-besu.dnp.dappnode.eth",
```
dsimog01, line 215: For the rest of the clients, you do not put "goerli-" here

dsimog01: I don't know if it is a mistake or not

pablomendezroyo: Thanks for the comment! That file is just for the mock backend so is not that important

PR #1190, opened 2022-11-09
Improving log colors for Stakers UI

dsimog01:
Added info, warning and danger colors for Stakers UI logs

### DIFF
--- a/packages/admin-ui/src/dappnode_styles.scss
+++ b/packages/admin-ui/src/dappnode_styles.scss
@@ -266,3 +266,8 @@ p.no-p-style:last-child {
     opacity: 0.7;
   }
 }
+
+.bottom-alert {
+  margin-top: 10px;
+  margin-bottom: 0px;
+}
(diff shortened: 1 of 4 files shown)


packages/admin-ui/src/dappnode_styles.scss
```diff
@@ -266,3 +266,8 @@ p.no-p-style:last-child {
     opacity: 0.7;
   }
 }
+
+.bottom-alert {
```
pablomendezroyo, line 270: styles applied in the dappnode_styles.scss should be generic and used in multiple places in the dappmanager. Also, to gain space at the top and bottom you could simply use the `<br />` HTML element

dsimog01: I tried to use the <br/> tag, but it did not work with the alert component. I would rather create a CSS file for this component, but I am not sure where to create it or if it is OK to do it like that.

dsimog01: Fixed on https://github.com/dappnode/DNP_DAPPMANAGER/pull/1190/commits/1342757c0019451e8ef5f2e5a2fafc46d3f3deae

pablomendezroyo (approved):

pablomendezroyo (approved):

PR #1201, opened 2022-11-11
Make ENS/DNP name the same for mainnet besu

alexpeterson91:
ENS for the package is different than the metadata file.
Not sure if it matters at all but thought if its one way in the definition of the package it should stay consistent.

## Context

## Approach


pablomendezroyo:
It should be the same, although is not critical

PR #1217, opened 2022-11-15
Update disclaimer

pablomendezroyo:
Update disclaimer for mev boost

### DIFF
--- a/packages/admin-ui/src/pages/stakers/data.ts
+++ b/packages/admin-ui/src/pages/stakers/data.ts
@@ -15,20 +15,18 @@ export const disclaimer = `## Terms of Use - DAppNode
 
 Effective as of Oct 14, 2020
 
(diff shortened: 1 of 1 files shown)


alexpeterson91 (commented):
Going on record here this is NOT sufficient, i spent quite a bit of time researching this and this was my assignment, hell I assigned it to myself in September, this isnt enough. I posted what was needed 5 places before the mainnet stakersUI was published as I was instructed. 
 I'm not just an American citizen, I am a former AML/BSA/KYC compliance officer who worked under the US treasury's crime reporting and enforcement arm, OFAC is major and jurisdiction goes way beyond the borders of the US. 
 You're asking for trouble with something you dont dare wanna mess with. @eduadiez we spoke about this and thought we had an understanding about how important this is not to play fast and loose with.  
 I'm really concerned and disappointed my updates to the disclaimers regarding this were ignored despite how much we discussed it and that my work was done the day before MEV-Boost was published, and still no one has reviewed them, i should have just committed them and not waited for any reviews for something only I am qualified to do.

PR #1234, opened 2022-11-18
enable gnosis for stakers UI

pablomendezroyo:
Enable gnosis chain in the stakers UI

### DIFF
--- a/packages/dappmanager/src/modules/stakerConfig/stakerParamsByNetwork.ts
+++ b/packages/dappmanager/src/modules/stakerConfig/stakerParamsByNetwork.ts
@@ -62,28 +62,28 @@ export function stakerParamsByNetwork<T extends Network>(
         execClients: [
           {
             dnpName: "nethermind-xdai.dnp.dappnode.eth",
-            minVersion: ""
+            minVersion: "1.0.17"
           }
         ],
         currentExecClient: db.executionClientGnosis.get(),
         consClients: [
           {
             dnpName: "gnosis-beacon-chain-prysm.dnp.dappnode.eth",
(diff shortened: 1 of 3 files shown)


3alpha (approved):

alexpeterson91:
Let's add erigon if we can. Client diversity is key on all chains. I started the packsge already.  But it's not done

packages/dappmanager/src/modules/stakerConfig/stakerParamsByNetwork.ts
```diff
@@ -65,7 +65,7 @@ export function stakerParamsByNetwork<T extends Network>(
             minVersion: ""
           }
         ],
```
alexpeterson91, line 67: Can we include logic for handling erigon? At least on another parallel branch to test it on.

alexpeterson91 (commented):
Consider Erigon in addition to nethermind for gnosis chssinsz

packages/dappmanager/src/modules/stakerConfig/stakerParamsByNetwork.ts
pablomendezroyo: Only tested, polished and published packages can be added in the stakers UI. However we can already take erigon into account when building the consensus clients for the stakers UI

alexpeterson91: of course, sorry i really meant we should keep it in mind currently by not just hardcoding nethermind xdai for everything and should just have a parallel branch with it included for testing erigon with the gnosis stakersUI

PR #1265, opened 2022-11-29
Update wording for Ethereum repository

pablomendezroyo:
Update wording for ethereum repository

### DIFF
--- a/packages/admin-ui/src/pages/system/actions.ts
+++ b/packages/admin-ui/src/pages/system/actions.ts
@@ -34,10 +34,10 @@ export const changeEthClientTarget = (
       await new Promise<void>(resolve =>
         confirm({
           title: `Remove ${prettyDnpName(prevTarget.execClient)}?`,
-          text: `Do you want to keep or remove your current Ethereum execution client? This action cannot be undone. Having more than one ETH client may cause your machine unexpedted behaviours`,
+          text: `Do you want to remove your current Execution Layer (EL) client? This action cannot be undone. You can keep the volume data to avoid resyncing from scratch next time you install the same EL client. Keeping the volume will NOT clear space in your hard drive.`,
           buttons: [
             {
-              label: "Keep",
+              label: "Keep node running",
               variant: "danger",
               onClick: () => {
                 deletePrevExecClient = false;
@@ -46,7 +46,7 @@ export const changeEthClientTarget = (
               }
             },
             {
-              label: "Remove without volumes",
+              label: "Remove and keep volumes",
               variant: "warning",
               onClick: () => {
                 deletePrevExecClient = true;
@@ -55,7 +55,7 @@ export const changeEthClientTarget = (
               }
             },
             {
-              label: "Remove with volumes",
+              label: "Remove and delete volumes",
               variant: "dappnode",
               onClick: () => {
                 deletePrevExecClient = true;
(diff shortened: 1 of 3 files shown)


packages/admin-ui/src/pages/system/actions.ts
```diff
@@ -46,7 +46,7 @@ export const changeEthClientTarget = (
               }
             },
             {
-              label: "Remove without volumes",
+              label: "Remove and keep volumes",
```
alexpeterson91, line 49:
```suggestion
              label: "Remove Package and retain volumes",
```

packages/admin-ui/src/pages/system/actions.ts
```diff
@@ -55,7 +55,7 @@ export const changeEthClientTarget = (
               }
             },
             {
-              label: "Remove with volumes",
+              label: "Remove and delete volumes",
```
alexpeterson91, line 58:
```suggestion
              label: "Remove Package and delete volumes",
```

packages/admin-ui/src/pages/system/actions.ts
```diff
@@ -34,10 +34,10 @@ export const changeEthClientTarget = (
       await new Promise<void>(resolve =>
         confirm({
           title: `Remove ${prettyDnpName(prevTarget.execClient)}?`,
-          text: `Do you want to keep or remove your current Ethereum execution client? This action cannot be undone. Having more than one ETH client may cause your machine unexpedted behaviours`,
+          text: `Do you want to remove your current Execution Layer (EL) client? This action cannot be undone. You can keep the volume data to avoid resyncing from scratch next time you install the same EL client. Keeping the volume will NOT clear space in your hard drive.`,
```
alexpeterson91, line 37:
```suggestion
          text: `Do you want to remove your current Execution Layer (EL) client? This action cannot be undone. You can keep the data volume to avoid needing to resync from scratch next time you install the same EL client. Keeping the data volume will NOT clear space in your hard drive.`,
```
 from personal experience testing the stakersUI with several ELs, ive found that theres an issue with well all ELs using snap sync in some way, where even if they are shut down properly they don't recover properly, for example geth is the worst offender in my experience it keeps me attesting most of the time but geth/ besu/nethermind/ (havn't had time switdhing to and from erigon given the very long sync times) i often find the need to resync from genesis. and for CLs its _Always_ faster to wipe the CL volume and start from checkpoint sync endpoint.   The issue for both in my experience appears to be one of the clients  forking off and  then forking the other off sometimes needing both EL and CL to have the volumes wiped.  am I alone in experiencing this at all? geth just always is trying to state heal for days or weeks on end, it attests but misses 1-2 a day, and stuff like pokt tells me that theres errors with the RPC for geth and wont relay ETH (which has always been my biggest request from the network, all other chains work for me save for ETH mainnet.  i often see chain reorg warns etc in the logs as well. not sure how to resolve this.  Also if the client is stopped is it a clean or unclean shutdown, since we should make sure it closes down the clients gracefully to save the DBs properly

packages/admin-ui/src/pages/system/actions.ts
```diff
@@ -88,7 +88,7 @@ export const changeEthClientTarget = (
               }
             },
             {
-              label: "Remove without volumes",
+              label: "Remove and keep volumes",
```
alexpeterson91, line 91:
```suggestion
              label: "Remove Package and retain volumes",
```

packages/admin-ui/src/pages/system/actions.ts
```diff
@@ -97,7 +97,7 @@ export const changeEthClientTarget = (
               }
             },
             {
-              label: "Remove with volumes",
+              label: "Remove and delete volumes",
```
alexpeterson91, line 100:
```suggestion
              label: "Remove Package and delete volumes",
```

alexpeterson91 (changes requested):
Couple minor things

packages/admin-ui/src/pages/system/actions.ts
pablomendezroyo: Looks to long for a button

packages/admin-ui/src/pages/system/actions.ts
pablomendezroyo: Looks too long for a button

packages/admin-ui/src/pages/system/actions.ts
pablomendezroyo: Looks too long for a button

packages/admin-ui/src/pages/system/actions.ts
pablomendezroyo: Looks too long for a button

PR #1285, opened 2022-12-02
Display progress percent with 2 fraction digits

Ucodia:
## Context

Currently chain synchronization progress is shown floored to the closest integer, as such during a refresh UI may jump from 99% to 100%. This is fine in cases where the time it takes to make 1% progress is linear. Though in reality chain synchronization progress may seriously slow down in the last few percent, as such the experience is often to see the UI stuck at 99% for a very long time.

The purpose of this PR is to allow showing 2 fractional digits on the progress bar as to provide more granular progress in order to provide more accurate feedback in the last few percent of chain synchronization.

## Approach

Use the `toFixed(2)` JavaScript number method to remove other fractional digits.

## Test instructions

Chain synchronization progress in both the chain card in the main dashboard and in the chain dropdown should show percentage as `xx.xx%` instead of `xx%`. You can stop a chain for a minute and restart in order to cause synchronization progress to show for that chain.

### DIFF
--- a/packages/admin-ui/src/components/navbar/dropdownMenus/BaseDropdown.tsx
+++ b/packages/admin-ui/src/components/navbar/dropdownMenus/BaseDropdown.tsx
@@ -10,12 +10,11 @@ type BubbleColor = "danger" | "warning" | "success" | "light";
 // Utilities
 
 const ProgressBarWrapper = ({ progress }: { progress: number }) => {
-  const progressPercent = Math.floor(100 * progress);
   return (
     <ProgressBar
-      now={progressPercent}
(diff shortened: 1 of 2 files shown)


Ucodia:
Oops, missed on the integration tests. Will check and update assertions where necessary.

Ucodia:
Failing tests are all for IPFS, unlikely to be related to this PR. Any clue?

dapplion (approved):
Looks good to me! Showing more decimals sounds sensible. However some protocols will still appear stuck at some high percentage due to how their sync algorithm works.

PR #1291, opened 2022-12-07
Improve `interceptGlobalEnvOnSet`

pablomendezroyo:
Improve the function `interceptGlobalEnvOnSet`:
- [x] Set types for db keys and values
- [x] Use those types in the fuction args
- [ ] Improve the checker for the `defaultEnvironment`


pablomendezroyo:
Not needed

PR #1294, opened 2022-12-08
Updated HTTPS api to handle external and internal connections

3alpha:
Updated HTTPS api so it is backwards compatible. We should have consensus on default behaviour.


pablomendezroyo:
What is the status of this PR?

3alpha:
We need to sync on this one. It updates API, but doesnt deal with UI.

pablomendezroyo:
@3alpha status?

PR #1308, opened 2022-12-15
Bum node version

pablomendezroyo:
Bump node version in workflows

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: "16.15.0"
+          node-version: "19.2.0"
       - run: yarn
       - run: yarn bootstrap
       - run: yarn generate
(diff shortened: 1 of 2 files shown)


alexpeterson91:
Yeah really want this to work. Been struggling to build a proper iso myself since unless I manually pull the builds from GH for the core packages or have build=true (which never works for me always fails for lack of dependencies but the script hides what's missing since it's well a cascading series of scripts to build an ISO. Just been trying to make my own ISOs for cloud install natively and another for nuc12s but even with the dappnode profile file updated it keeps pulling older versions like dappmanager 0.50. Also I'm curious why is there no tag for core 2.63? That's what I'm on.

.github/workflows/main.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v1
         with:
-          node-version: "16.15.0"
+          node-version: "19.2.0"
```
dapplion, line 69: Odd NodeJS versions are not stable, why not use 18?

pablomendezroyo:
Closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1317

PR #1318, opened 2022-12-21
update wording

alexpeterson91:
update wording
Have more to do will be for next core release.

## Context

## Approach


pablomendezroyo (approved):

alexpeterson91:
i still need to get back to this but im just backed up, might as well merge whats done already

PR #1319, opened 2022-12-21
fix more links, remove gitcoin references.

alexpeterson91:
fix more links, remove gitcoin references.

## Context

## Approach

### DIFF
--- a/packages/admin-ui/src/pages/community/components/TreasuryActions.tsx
+++ b/packages/admin-ui/src/pages/community/components/TreasuryActions.tsx
@@ -1,14 +1,11 @@
 import Button from "components/Button";
 import React from "react";
-import { mediumTreasuryUrl, explorerTreasuryUrl } from "params";
+import explorerTreasuryUrl from "params";
 import newTabProps from "utils/newTabProps";
 
 export default function TreasuryActions() {
   return (
     <>
-      <Button href={mediumTreasuryUrl} {...newTabProps}>
-        Read more
-      </Button>
       <Button href={explorerTreasuryUrl} {...newTabProps}>
         Check my ranking
       </Button>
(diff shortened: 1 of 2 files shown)


alexpeterson91:
The links updated. in Params are more important than removing gitcoin, but we do need to remove all references to our old gitcoin info which i thought i did already a few months ago, but not all of it apparently specifically the link to the medium article explaining the old community treasury and Panvala gitcoin cycle, left the source cred explorer though

alexpeterson91:
added P0 since itd be great to finish merging all the updated urls before publishing the new core, this is my last PR for the dappmanager's docs updates for now.

pablomendezroyo (approved):

dsimog01 (approved):

packages/admin-ui/src/pages/community/components/TreasuryActions.tsx
```diff
@@ -1,14 +1,11 @@
 import Button from "components/Button";
 import React from "react";
-import { mediumTreasuryUrl, explorerTreasuryUrl } from "params";
+import {explorerTreasuryUrl} from "params";
 import newTabProps from "utils/newTabProps";
 
 export default function TreasuryActions() {
   return (
     <>
-      <Button href={mediumTreasuryUrl} {...newTabProps}>
```
pablomendezroyo, line 9: Why removing the treasury?

alexpeterson91: Discus during/after daily? Not worth to write it out right now.

alexpeterson91: agreed we should probably just remove the entire section for the treasury, link already removed is way outdated and doesnt work like that, the remaining explorer link does work ans shows the sourcecred dashboard to show stats, but it doesnt pay. out, hasnt since April, and it doesnt allow new applicants on discord since the channels and bots were removed/shut disabled for months now since the last revamp.  we have a new way to track all user contributions with the Statbot on discord and I think is a far better thing to do now no need to deal with the #ct-addresses channel doesnt even exist anymore and just message the largest PAN holders to  withdraw now if they so desire, PAN fell massively in value as well. making it much less relevant, let me know if you agree to just remove the whole section and ill add those commits here

pablomendezroyo (approved):

PR #1327, opened 2022-12-22
Moved repository tab to left menu (navbar)

dsimog01:
The Repository view is moved from System>Repository to an own tab. Moreover, the view is divided into 2 submenus, one for IPFS and the other for Ethereum


pablomendezroyo (changes requested):
Would be great to move `add-ipfs-peer` also to repository > IPFS

PR #1336, opened 2022-12-27
Set timeout for compose calls if same dcPath

pablomendezroyo:
To avoid doing multiple compose call to the same compose file, that may lead to errors, perform a timeout whenever the dcpath is the same as the previous one


dapplion:
@pablomendezroyo please just build a queue with concurrency of 1

dapplion:
@pablomendezroyo please just build a queue with concurrency of 1

PR #1337, opened 2022-12-27
Use ESM modules

pablomendezroyo:
Use ESM modules in dappmanager and admin-ui

- dappmanager:
  - Bump ipfs-http api library
  - In dappmanager `tsconfig.json` use module EsNext and target EsNext to avoid issues in newest ipfs http library
  - Set `"type": "module"` in dappmanager `package.json` to be able to import src code from test folder
  - Bump `webpack-cli` library and use set extension `.cjs` in config file
  - Use `lodash-es` instead of `lodash`


pablomendezroyo:
closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1341

PR #1340, opened 2022-12-29
Update ipfs http client library

pablomendezroyo:
Update ipfs http client library. To make it work with webpack compilation, it was needed to set `module` to `ESNext` in the `tsconfig.json`

The reason is that ipfs http client library now is only ESM compatible (from release 57)


pablomendezroyo:
closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1341

PR #1380, opened 2023-02-15
Implement default fee recipient

pablomendezroyo:
Set `FEE_RECIPIENT_<NETWORK>` global env var. The compose env `FEE_RECIPIENT_ADDRESS` value will be retrieved from the Consensus Client selected, if not possible to retrieve the value the global env `FEE_RECIPIENT_<NETWORK>` will be set to an empty string, in order to avoid triggering the migration again.

### DIFF
--- a/packages/dappmanager/src/modules/migrations/setDefaultFeeRecipient.ts
+++ b/packages/dappmanager/src/modules/migrations/setDefaultFeeRecipient.ts
@@ -18,6 +18,7 @@ export async function setDefaultFeeRecipient(): Promise<void> {
   for (const network of networks) {
     const envVarName = `FEE_RECIPIENT_ADDRESS`;
     let consensusClientDnpName: ConsensusClient<Network> | null;
+    let envVarValue = "";
 
     switch (network) {
       case "prater":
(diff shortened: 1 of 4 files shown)


packages/dappmanager/src/modules/migrations/setDefaultFeeRecipient.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export async function setDefaultFeeRecipient(): Promise<void> {
+  const networks: Network[] = [];
+
+  if (db.feeRecipientPrater.get() === null) networks.push("prater");
+  if (db.feeRecipientGnosis.get() === null) networks.push("gnosis");
+  if (db.feeRecipientMainnet.get() === null) networks.push("mainnet");
+
+  for (const network of networks) {
+    const envVarName = `FEE_RECIPIENT_ADDRESS`;
```
dsimog01, line 19: Why is this FEE_RECIPIENT_ADDRESS? Above you used FEE_RECIPIENT_<NETWORK> (no "ADDRESS"). Can this be problematic?

pablomendezroyo: It is the env var name used in clients compose files (i.e https://github.com/dappnode/DAppNodePackage-teku-prater/blob/6da01836c0dc395e90e25381d36913d3802c6558/docker-compose.yml#L13)

PR #1416, opened 2023-03-15
Update light_dark.scss

Kingpin98k:
I changed the color of alert texts to white, for all types of selectors in dark mode.

## Context

This pull request addresses an issue where the text color of alert banners is difficult to read in dark mode. The issue occurred because the alert text color was set to black for all color schemes, including dark mode. As a result, the text would blend into the background in dark mode, making it difficult to read.

To fix this issue, this pull request updates the CSS for alert banners so that the text color is white when the user is in dark mode. This ensures that the alert banners are easily readable in all color schemes.

## Approach

To fix the issue of unreadable alert banners in dark mode, this pull request updates the CSS for alert banners to include a media query that targets the prefers-color-scheme property of the device. The new CSS rule ensures that when the user is in dark mode, the text color of alert banners is set to white. This change makes the text easily readable in dark mode, while still preserving the original black color in light mode.

Test Instructions :

Prerequisites ->

A web page with alert components that use the .alert, .alert-secondary, and .alert-info CSS classes.
A device with a browser that supports CSS variables and media queries.

Test Steps ->

. Ensure that the website is in light mode.
. Check that the text color of all alert types is black.
. Change the preferred color scheme of your device to dark mode.
. Reload the web page.
. Check that the text color of all alert types is white.
. Change the preferred color scheme of your device back to light mode.
. Reload the web page.
. Check that the text color of all alert types is black again.

Expected Results ->

. In light mode, the text color of all alert types should be black.
. In dark mode, the text color of all alert types should be white.
. Switching between light and dark mode should cause the text color of the alert types to change accordingly.


pablomendezroyo (changes requested):
Thanks for the contribution!

I ve seen the results and, although its better, still not legible enough. Probably because of the background color (yellow). 

Ive tried with the color ` var(--color-dark-quaternarytext);` which looks better. Could you try it and commit the suggestion?

pablomendezroyo (approved):

PR #1426, opened 2023-03-28
Add flyway service checker

pablomendezroyo:
Add flyway service checker for web3signer package.

### DIFF
--- a/packages/dappmanager/src/modules/stakerConfig/utils.ts
+++ b/packages/dappmanager/src/modules/stakerConfig/utils.ts
@@ -178,15 +178,25 @@ export async function updateMevBoostEnv<T extends Network>({
   }
 }
 
+/**
+ * Returns true if the package is running or false if not
+ * For web3signer, it does not take into account the container "flyway" which may not be running
+ */
 export function getIsRunning(
   { dnpName }: { dnpName: string },
   dnpList: InstalledPackageData[]
 ): boolean {
-  return (
-    dnpList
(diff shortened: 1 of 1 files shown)


packages/dappmanager/src/modules/stakerConfig/utils.ts
```diff
@@ -178,15 +178,25 @@ export async function updateMevBoostEnv<T extends Network>({
...
 export function getIsRunning(
   { dnpName }: { dnpName: string },
   dnpList: InstalledPackageData[]
 ): boolean {
-  return (
-    dnpList
-      .find(dnp => dnp.dnpName === dnpName)
-      ?.containers.every(c => c.running) ?? false
-  );
+  const flywayServiceName = "flyway";
```
dsimog01, line 189: Maybe this way it can be more readable and reduce complexity?

```export function getIsRunning(
  { dnpName }: { dnpName: string },
  dnpList: InstalledPackageData[]
): boolean {
  const flywayServiceName = "flyway";
  const isSigner = dnpName.includes("web3signer");
  const dnp = dnpList.find(dnp => dnp.dnpName === dnpName);

  if (!dnp) return false;

  let nonFlywayContainers: ContainerData[] | undefined;

  if (isSigner) {
    nonFlywayContainers = dnp.containers?.filter(c => c.serviceName !== flywayServiceName);
  } else {
    nonFlywayContainers = dnp.containers;
  }

    return nonFlywayContainers?.every(c => c.running) ?? false;
  }
```

PR #1432, opened 2023-03-29
Bum dappnodesdk dependency

pablomendezroyo:
Bum dappnodesdk dependency to 0.2.82 and modify the way how is imported.

Importing modules (`exports` in the `package.json`) is not currently working in typescript


pablomendezroyo:
Closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1438

PR #1437, opened 2023-04-05
Definitive lodestar compatible versions

dsimog01:
All clients and signers are now compatible with Lodestar, so the versions can be set up definitely now


dapplion (approved):
:clap:

PR #1443, opened 2023-04-11
Improve `tsconfig.json`

dsimog01:
Included a root `tsconfig.json` to simplify the organization and removed unnecessary options


Dockerfile
```diff
@@ -50,6 +50,7 @@ RUN yarn --frozen-lockfile --non-interactive --ignore-optional
 # Build common
 WORKDIR /app/packages/common/
 COPY packages/common/ .
+COPY tsconfig.json /app
```
pablomendezroyo, line 53: This step does not belong to the build common so I would move it a few lines up

pablomendezroyo (changes requested):

Dockerfile
dsimog01: True, fixing it

pablomendezroyo (approved):

PR #1444, opened 2023-04-11
Added ETC Cooperative key

dsimog01:
Added key to dappmanager so that whenever the ETC Cooperative publish packages it shows the `ETC Cooperative` as author of the package.


pablomendezroyo (approved):

dsimog01:
Installed the built package and it shows the key properly

Issue #1427, opened 2023-03-28
Set `useCheckpointSync` as true by default when selecting a conensus client in the stakers


pablomendezroyo:
When a user selects a new consensus client in the stakers, set the useCheckpointSync to true by default

pablomendezroyo:
Closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1455

PR #1455, opened 2023-04-17
Set `useCheckpointSync` to true for all concensus clients

Marketen:
## Context

PR related to issue #1427. Some concensus clients had **_useCheckpointSync_** defined to false by default. Others didn't have it defined (when **_useCheckpointSync_** its not defined properly in each concensus client, its value is set to false, this does not change in this PR) https://github.com/dappnode/DNP_DAPPMANAGER/blob/f3f654426ba28e2b770236fcae3a25e52b522339/packages/admin-ui/src/pages/stakers/components/columns/ConsensusClient.tsx#L27

## Approach

Each concensus clients now has its **_useCheckpointSync_** variable to "true" by default. 

## Test instructions

Check that sync is set to true in "Stakers" frontend by default
image

### DIFF
--- a/packages/admin-ui/src/__mock-backend__/stakerConfig.ts
+++ b/packages/admin-ui/src/__mock-backend__/stakerConfig.ts
@@ -186,7 +186,8 @@ export const stakerConfig: Pick<
                   shortDescription: "Lighthouse consensus client",
                   version: "0.1.0"
                 }
-              }
+              },
+              useCheckpointSync: true
             },
             {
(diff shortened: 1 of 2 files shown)


pablomendezroyo (changes requested):
These changes do not reflect to use the checksync in the consensus client. This is the mock used to generate "fake" backend data to simulate the frontend. You must have a look at the file `StakerNetwork`

pablomendezroyo (approved):

PR #1457, opened 2023-04-18
Fix import for monero

dsimog01:
Monero chain card was showing `Daemon is not a constructor`. This issue is fixed by updating the import of `monero-rpc` library.
image

### DIFF
--- a/packages/dappmanager/src/modules/chains/drivers/monero.ts
+++ b/packages/dappmanager/src/modules/chains/drivers/monero.ts
@@ -1,3 +1,4 @@
+// @ts-ignore
 import { Daemon } from "monero-rpc";
 import { InstalledPackageData } from "@dappnode/common";
 import { getPrivateNetworkAlias } from "../../../domains.js";
--- a/packages/dappmanager/src/modules/chains/drivers/monero.ts
+++ b/packages/dappmanager/src/modules/chains/drivers/monero.ts
@@ -1,5 +1,5 @@
 // @ts-ignore
-import Daemon from "monero-rpc";
+import { Daemon } from "monero-rpc";
 import { InstalledPackageData } from "@dappnode/common";
 import { getPrivateNetworkAlias } from "../../../domains.js";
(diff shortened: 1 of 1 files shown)


alexpeterson91:
theres a handful of other cards that need updating due to similar reasons deprecated methods and other api calls, a few are port related too, ill catch up with you on discord tomorrow about them .
None have the same issue but i know whats wrong with each of them. Just would like some help fixing the DAPPMANAGER side for a most of 'em.  One is on the package side but i tried changing values a few different ways and nothing worked so i think the code block for ETH RPC calls for card statuses so that its not hardcoded to always be `8545` rather the port taken from the end of the URL in each package manifest for `api` `endpoint` or `queryingApi` (all are interpreted as the same by the DAPPMANAGER when displaying `Querying-API` on the info tab of the package.)

packages/dappmanager/src/modules/chains/drivers/monero.ts
```diff
@@ -1,5 +1,5 @@
 // @ts-ignore
-import Daemon from "monero-rpc";
+import { Daemon } from "monero-rpc";
```
pablomendezroyo, line 2: Looks like now you can delete the `// @ts-ignore`

pablomendezroyo (changes requested):

packages/dappmanager/src/modules/chains/drivers/monero.ts
pablomendezroyo: bump @dsimog01

pablomendezroyo (approved):

packages/dappmanager/src/modules/chains/drivers/monero.ts
dsimog01: ts-ignore is still needed. The build fails and there is no `@types/monero-rpc`

pablomendezroyo (approved):

PR #1466, opened 2023-05-03
Set concurrency 1 in staker global envs docker calls

pablomendezroyo:
Set concurrency 1 in staker global envs docker calls to ensure the actions are executed one after the other


packages/dappmanager/src/modules/stakerConfig/setStakerConfig.ts
```diff
@@ -175,7 +175,7 @@ export async function setStakerConfig<T extends Network>({
     mevBoost,
     targetMevBoost: stakerConfig.mevBoost,
     currentMevBoostPkg: pkgs.find(pkg => pkg.dnpName === mevBoost)
-  }).then(() => setMevBoostOnDb(stakerConfig.network, stakerConfig.mevBoost?.dnpName));
+  }).then(async () => await setMevBoostOnDb(stakerConfig.network, stakerConfig.mevBoost?.dnpName));
```
dsimog01, line 178: Same as above

dsimog01 (changes requested):

PR #1481, opened 2023-05-15
Bump dappnode types dependency and consider monoservice

pablomendezroyo:
Bump dappnode types dependency and consider monoservice for tagging images and setting compose aliases


dsimog01 (approved):

dsimog01 (changes requested):
Actions are failing

PR #1494, opened 2023-06-14
Deprecate docker update and host update

pablomendezroyo:
Deprecate docker update and host update. These functionalities are critical and must be reviewed and updated:
- Host update should be done through core package
- Docker update should be done through core package and they must be switched from package to apm repo installation method


dsimog01 (changes requested):
Action is failing (not able to build)

PR #1503, opened 2023-07-06
Use await all in staker

pablomendezroyo:
Use await all in stakers configuration

### DIFF
--- a/packages/dappmanager/src/modules/stakerConfig/set/setStakerConfig.ts
+++ b/packages/dappmanager/src/modules/stakerConfig/set/setStakerConfig.ts
@@ -69,14 +69,14 @@ export async function setStakerConfig<T extends Network>({
   await setFeeRecipientOnDb(network, feeRecipient || undefined);
   await Promise.all([
     // EXECUTION CLIENT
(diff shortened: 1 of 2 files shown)


packages/dappmanager/src/modules/stakerConfig/set/setStakerConfig.ts
```diff
@@ -65,42 +65,39 @@ export async function setStakerConfig<T extends Network>({
...
-  // MEV BOOST
-  await setMevBoost({
-    network,
-    mevBoost: compatibleMevBoost?.dnpName,
-    targetMevBoost: mevBoost,
-    currentMevBoostPkg
-  });
+  await Promise.all([
+    // EXECUTION CLIENT
+    await setExecutionClient<T>({
```
dsimog01, line 72: Do you need to use `await` inside of `Promise.all`?

pablomendezroyo: Nice catch!

dsimog01 (approved):

PR #1506, opened 2023-07-07
Fix dev mode

pablomendezroyo:
Fix dev mode and improves its efficiency by executing the installs and builds in docker build stage


dsimog01 (approved):

pablomendezroyo:
ref https://github.com/TypeStrong/ts-node/issues/1997

PR #1509, opened 2023-07-10
Check is running in set staker config

pablomendezroyo:
Check is running in set staker config to avoid calling unnecessary `docker compose`calls


pablomendezroyo:
Wont implement

PR #1521, opened 2023-07-12
Migrate from react-bootstrap to mui

pablomendezroyo:
Migrate from react-bootstrap to material UI


pablomendezroyo:
wont implement

PR #1532, opened 2023-07-17
Update DB on pkg remove

dsimog01:
On pkg remove, the DB should be updated if that pkg was selected in the stakers tab


3alpha:
Is there reason no to handle that here: https://github.com/dappnode/DNP_DAPPMANAGER/blob/develop/packages/dappmanager/src/calls/packageRemove.ts?

packages/dappmanager/src/daemons/stakerDbUpdate/index.ts
```diff
@@ -0,0 +1,59 @@
...
+import { mevBoostMainnet, mevBoostPrater, stakerPkgs } from "@dappnode/types";
+import * as db from "../../db/index.js";
+import { logger } from "ethers";
+
+async function removeStakerPkgsFromDbIfSelected({
+  dnpNames
+}: {
+  dnpNames?: string[];
+}): Promise<void> {
+  if (!dnpNames) return;
```
pablomendezroyo, line 11: this checker will not work as you expect. The `if(!...)` does not check for arrays with length 0. You must check for `if(dnpNames.length === 0)

packages/dappmanager/src/daemons/stakerDbUpdate/index.ts
```diff
@@ -0,0 +1,59 @@
...
+        break;
+      case "mev-boost-goerli.dnp.dappnode.eth":
+        db.mevBoostPrater.set(false);
+        break;
+      default:
+        break;
+    }
+  });
+
+  logger.info(`Removed clients/mev-boost ${dnpNames} from main DB`);
```
pablomendezroyo, line 43: Too ambiguous and not precise. Not all packages might belong to the main DB. Either remove it or improve it

packages/dappmanager/src/daemons/stakerDbUpdate/index.ts
```diff
@@ -0,0 +1,59 @@
...
+      case db.consensusClientMainnet.get():
+        db.consensusClientMainnet.set(undefined);
+        break;
+      case db.consensusClientGnosis.get():
+        db.consensusClientGnosis.set(undefined);
+        break;
+      case db.consensusClientPrater.get():
+        db.consensusClientPrater.set(undefined);
+        break;
+      case "mev-boost.dnp.dappnode.eth":
```
pablomendezroyo, line 32: Use the dnpName provided by the library @dappnode/types

pablomendezroyo (changes requested):

PR #1550, opened 2023-08-03
Move and rename getEndpointConfig()

Marketen:
## Context

getEndpointConfig() method was missplaced in the code and had a non intuitive name
## Approach

renamed and moved the method to "packages/dappmanager/src/modules/docker/api/network.ts" a file related to docker API calls to docker network
## Test instructions

on _/packages/dappmanage_ dir, run `yarn run test:int` to execute unit test related to this new renamed method `getDnCoreNetworkContainerAliases(containerName)`. This will verify that when calling this method with the dappmanager container, it returns a network (dn_core) confuguration and it is not null
-->


pablomendezroyo:
@Marketen int test failling 
```
1) DNP lifecycle
    Test stopping and removing
       DNP should be running:

```

pablomendezroyo (approved):

pablomendezroyo:
great work!

PR #1556, opened 2023-08-11
Integrate Ethical metrics in dappmanager

pablomendezroyo:
Integrate Ethical metrics in dappmanager

- [x] Add ethical metrics as new welcome feature
- [x] Add ethical metrics as new notifications system in **system > notifications**
- [x] Add ethical metrics daemon that on enabled, install ethical metrics and register. On disable, unregister
- [x] Add migration to fetch email and set it to the db


dsimog01:
Definition for the integration in Dappmanager: 

1. Creation of a new global env called `_DAPPNODE_GLOBAL_USER_EMAIL`
2. Addition of that global env to Ethical Metrics package
3. Migration on core install: Global env must be written with the current value of the email set in Ethical Metrics (or empty if it is not installed)
4. Add a box inside System>Notifications were Ethical Metrics management can be performed. User should be able to: See email, instance and register status, register, unregister and update email (Show banner that suggest installation if EM is not installed)
5. Add Ethical Metrics set up to onboarding process

PR #1561, opened 2023-08-23
Allow empty fee recipient in stakers

dsimog01:
The fee recipient field in the stakers becomes empty after refreshing, except if the fee recipient is already stored in the `maindb.json`. The DB is not updated when hitting "Appy changes". For Lukso this fix is necessary, as there is no user whose DB includes the fee recipient for LUKSO


dsimog01:
Closing this PR as https://github.com/dappnode/DNP_DAPPMANAGER/pull/1562 has been merged

PR #1563, opened 2023-08-25
Fix LUKSO checkpoint

dsimog01:
LUKSO checkpoint is not supported by Dappnode, it is supported by the LUKSO team


dsimog01:
Identation has been modifed. Closing this PR to fix it in a new one

PR #1582, opened 2023-09-13
Randomize Ethical Metrics cron

dsimog01:
Ethical Metrics cron interval should be randomized in order to reduce the number of simultaneous calls to the Ethical Metrics server


packages/dappmanager/src/daemons/ethicalMetrics/checkEthicalMetricsStatus.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+    } catch (e) {
+        logs.error("Error on ethical metrics check", e);
+    }
+}
+
+async function ensureAllContainersRunning(pkg: InstalledPackageData): Promise<void> {
+    // If the package is already installed, ensure it's running
+    if (pkg.containers.some(c => !c.running))
+        await dockerComposeUpPackage(
```
pablomendezroyo, line 71: could you use here a more targeted action such as 
```
for (const container of pkg.containers) {
  if(!container.running) {
    await dockerContainerStart(container.containerName)
    }
}
```

pablomendezroyo (changes requested):

PR #1588, opened 2023-09-15
DNS resolution managed by Docker DNS server

dsimog01:
Domain resolution should be made by the native docker DNS server, so that we do not need to keep a list in the bind package. This package should only be responsible for receiving the external requests (from a machine connected to the dappnode) and redirect ENS or DNS requests to the specific resolver.

This PR should avoid the internal DNS requests to go through the bind package. Also, `dns` statements in all the compose files inside dappmanager should be gone.

To test this, we should make sure that both installed packages and new packages resolve properly every domain


dappnodedev:
Overridden by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1645

PR #1617, opened 2023-09-26
Add determineNetworkAlias function to simplify aliases creation

pablomendezroyo:
Add `determineNetworkAlias` function to simplify aliases creation


packages/dappmanager/src/domains.ts
```diff
@@ -1,53 +1,75 @@
...
+ * @returns the root private network alias. It
  */
-export function shortUniqueDappnodeEns(dnpName: string): string {
-  for (const s of [".dnp.dappnode.eth", ".dappnode.eth", ".eth"])
-    if (dnpName.endsWith(s)) dnpName = dnpName.slice(0, -s.length);
-  return stripCharacters(dnpName);
-}
-
-export type ContainerNames = { serviceName: string; dnpName: string };
+export function determineNetworkAlias({
```
dappnodedev, line 28: I would call this function `buildNetworkAlias` or `generateNetworkAlias`, as determine is not very accurate and could look like it sets the value somewhere but returns void

dappnodedev (changes requested):

packages/dappmanager/src/domains.ts
```diff
@@ -56,10 +78,39 @@ export function getPrivateNetworkAliases(
...
+ * - Strip container prefix
+ * - Strip .dappnode, .eth, .dnp
+ * - Strip "_"
+ *
+ * @param name "bitcoin.dnp.dappnode.eth"
+ * @returns "bitcoin"
+ * - "bitcoin.dnp.dappnode.eth" > "bitcoin"
+ * - "other.public.dappnode.eth" > "other.public"
+ */
+function getShortDnpName(dnpName: string): string {
```
dappnodedev, line 112: I think I do not fully understand this function. Why does it return `other.public` if dnp is `other.public.dappnode.eth`? I do not see where this is being done

packages/dappmanager/src/modules/chains/drivers/bitcoin.ts
```diff
@@ -35,7 +35,14 @@ export async function bitcoin(
 ): Promise<ChainDataResult> {
   const container = dnp.containers[0];
   if (!container) throw Error("no container");
-  const containerDomain = getPrivateNetworkAlias(container);
+
+  const {dnpName,serviceName} = container;
```
dappnodedev, line 39: In order to avoid having to duplicate this code in every driver, I would add a function called `build/generateNetworkAliasByPkg({service: string, pkg: InstalledPackageData})`

It would check if the package is monoservice or the specified service is main to set the value of `isMainOrMonoservice`

Then, it would call `build/generateNetworkAlias` to get an available alias

packages/dappmanager/src/modules/ethClient/apiUrl.ts
```diff
@@ -14,12 +14,14 @@ export function getEthExecClientApiUrl(dnpName: string, port = 8545): string {
    * domain = "bitcoin.dappnode", "other.public.dappnode"
    * ```
    */
-  const domain = getPrivateNetworkAlias({
-    dnpName: dnpName,
-    serviceName: dnpName
+  
+  const containerDomain = determineNetworkAlias({
```
dappnodedev, line 18: This does not look 100% reliable. What if you call this function with a dnp name that is a multiservice with no main service?

packages/dappmanager/src/modules/ethClient/apiUrl.ts
```diff
@@ -38,19 +40,21 @@ export async function getEthConsClientApiUrl(dnpName: string): Promise<string> {
     dnp.chain.serviceName
   ) {
     port = dnp.chain.portNumber;
-    domain = getPrivateNetworkAlias({
+    domain = determineNetworkAlias({
```
dappnodedev, line 43: You could also call here `build/generateNetworkAliasByPkg({service: string, pkg: InstalledPackageData})`

dappnodedev: `dnp.chain.serviceName` does not seem 100% reliable either

packages/dappmanager/src/domains.ts
```diff
@@ -1,53 +1,75 @@
...
 
-export function stripCharacters(s: string): string {
-  return s.replace(RegExp("_", "g"), "");
-}
-
-export function stripBadDomainChars(s: string): string {
-  // eslint-disable-next-line no-useless-escape
-  return s.replace(/[^a-zA-Z\-]+/g, "-");
-}
+export type ContainerNames = { serviceName: string; dnpName: string };
```
dappnodedev, line 4: Is this being used without `isMainOrMonoservice` somewhere? You might want to add it to this type

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
-    endpointConfig &&
-      endpointConfig.Aliases &&
-      Array.isArray(endpointConfig.Aliases) &&
-      endpointConfig.Aliases.includes(alias)
-  );
+export async function addAliasToGivenContainers(containers: PackageContainer[]): Promise<void> {
+  for (const container of containers) {
+
+    const isMainOrMonoService = container.isMain ?? false; // Set a default value of false if isMain is undefined
+    const service = {
```
dappnodedev, line 48: Here you could define the type

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
-      }
-      logs.info(`Added alias to running container ${container.containerName}`);
-    } catch (e) {
-      logs.error(`Error adding alias to container ${containerName}`, e);
-    }
+  try {
+    const containers = await listContainers();
+    await addAliasToGivenContainers(containers);
+  } catch (error) {
+    logs.error('Error adding alias to running containers:', error);
```
dappnodedev, line 40: Make sure you throw verbose errors where they could occur

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
 
-  // 4. Ensure compose file version 3.5
+  // Gets the old network "network" from the service compose file. This is an old network that is to be removed".
+  // If it is already removed, return null
+  const serviceNetwork = serviceNetworks[params.DNP_PRIVATE_NETWORK_NAME_FROM_CORE] ?? null;
+
+  // Return if migration was done, compose is already updated
+  if (isComposeNetworkAndAliasMigrated(composeNetwork, serviceNetwork, compose.compose.version, newAliases)) return
+  
+  // Ensure/update compose file version 3.5
```
dappnodedev, line 105: Are we sure this is the only format that we allow? Can someone install a package with a different format?

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
 
-  // 5. Add network and alias
-  if (composeNetwork || serviceNetwork)
-    // composeNetwork and serviceNetwork might be null and have different values (eitherway it should be the same)
-    // Only remove network if exists
-    composeService.removeNetwork(params.DNP_PRIVATE_NETWORK_NAME_FROM_CORE);
+  // This tries to remove the old network called "network" from the container in the compose file
+  // It is only done if network "network" exists or the new network "dncore_network" exists
+  if (composeNetwork || serviceNetwork) {
+    compose.services()[container.serviceName].removeNetwork(params.DNP_PRIVATE_NETWORK_NAME_FROM_CORE);
```
dappnodedev, line 114: Does this throw an error if `dncore_network` exists but `network` not?

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
+      dnpName: container.dnpName,
+      isMain: isMainOrMonoService, // Add the isMain property here
+    };
+    const aliases = getPrivateNetworkAliases(service)
+    
+    // Adds aliases to the compose file that generated the container
+    migrateCoreNetworkAndAliasInCompose(container, aliases);
+
+    // Adds aliases to the container network
+    for (const alias of aliases) {
```
dappnodedev, line 59: You could try to avoid the loop here.

1. Get the current alias in config
2. Compare those to the new ones and leave only those to add
3. Add them at the same time of possible

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
+    ...currentEndpointConfig,
+    Aliases: [...(currentEndpointConfig?.Aliases || []), alias]
+  };
+}
+
+async function updateContainerNetwork(networkName: string, container: any, endpointConfig: Partial<Dockerode.NetworkInfo>): Promise<void> {
+  const containerName = container.containerName;
+
+  // Wifi and VPN containers need a refresh connect due to their own network configuration
+  if (containerName === params.vpnContainerName || containerName === params.wifiContainerName) {
```
dappnodedev, line 145: Are we sure only those? Are both wireguard and OpenVPN considered here?

packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -26,140 +26,176 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
+ * @returns boolean
+ */
+function hasAlias(
+  endpointConfig: Dockerode.NetworkInfo | null,
+  alias: string
+): boolean {
+  return Boolean(
+    endpointConfig &&
+    endpointConfig.Aliases &&
+    Array.isArray(endpointConfig.Aliases) &&
```
dappnodedev, line 166: Could it be just a string and not an array? Likely not, but just in case

packages/dappmanager/src/modules/optimism/getOptimismConfig.ts
```diff
@@ -26,41 +26,48 @@ export async function getOptimismConfig(): Promise<OptimismConfigGet> {
 
     return {
```
dappnodedev, line 27: This should not be here, you need to rebase carefully

packages/dappmanager/test/integration/migrations/aliasMigration.test.int.ts
```diff
@@ -0,0 +1,255 @@
...
+// - monoContainer            --> container comming from a docker-compose with only 1 service. dnp
+// - monoContainerPublic      --> container comming from a docker-compose with only 1 service. public
+// - containerMain            --> container comming from a docker-compose with 2 services, it is maked as the main one. dnp
+// - containerNotMain         --> contaienr comming from a docker-compose with 2 services, it is not marked as the main one. dnp
+// - containerMainPublic      --> container comming from a docker-compose with 2 services, it is maked as the main one. public
+// - containerNotMainPublic   --> contaienr comming from a docker-compose with 2 services, it is not marked as the main one. public
+const containers = [containerMain, containerNotMain, monoContainer, containerMainPublic, containerNotMainPublic, monoContainerPublic];
+
+const CONTAINER_COMPOSE = `
+version: '3.4'
```
dappnodedev, line 86: Here you set version 3.4, but above you migrate to 3.5 right? Why is this? I might be missing something

dappnodedev (changes requested):

packages/dappmanager/src/domains.ts
dappnodedev: Done

packages/dappmanager/src/domains.ts
dappnodedev: It is being used without it, we'd better leave this

PR #1623, opened 2023-09-29
Fix OP Erigon not showing peers

dappnodedev:
OP Erigon and OP geth include the flag `--nodiscover` so in the case of OP Erigon this is probably the cause that `net_peerCount` returns an error. This error should not prevent the user from seeing whether the client is synced or not

Screenshot from 2023-09-29 09-56-03


dappnodedev:
After the fix: 
image

PR #1589, opened 2023-09-15
Update docker alias migrations

Marketen:

PR #1632, opened 2023-10-03
Docker alias migration

dappnodedev:
This PR overrides both https://github.com/dappnode/DNP_DAPPMANAGER/pull/1589 and https://github.com/dappnode/DNP_DAPPMANAGER/pull/1617

It includes the changes related to docker alias migration after the addition of the docker compose module in the dappmanager


packages/dappmanager/src/modules/migrations/addAliasToRunningContainers.ts
```diff
@@ -28,140 +28,223 @@ const dncoreNetworkName = params.DNP_PRIVATE_NETWORK_NAME;
...
 
-    try {
-      // Info from docker inspect and compose file might be not-syncrhnonyzed
-      // So this function must be before the check hasAlias()
-      migrateCoreNetworkAndAliasInCompose(container, alias);
+    // Adds aliases to the compose file that generated the container
+    migrateCoreNetworkAndAliasInCompose(container, aliases);
 
+    // Adds aliases to the container network
+    for (const alias of aliases) {
```
Marketen, line 62: We could try to avoid this loop, was pending from the other PR

Marketen: done

PR #1639, opened 2023-10-10
Bump schemas dependency

pablomendezroyo:
Bump dappnode schemas dependency


pablomendezroyo:
Schemas is an internal dependency

PR #1640, opened 2023-10-10
Fix `fullnode.dappnode` alias

dappnodedev:
`fullnode.dappnode` domain should point to the mainnet execution selected by the user, if any

Also, each network (prater, gnosis and lukso) has its own fullnode domain, following this pattern: `<network>.fullnode.dappnode`

### DIFF
--- a/packages/dappmanager/src/modules/ethClient/ethereumClient.ts
+++ b/packages/dappmanager/src/modules/ethClient/ethereumClient.ts
@@ -3,7 +3,7 @@ import { Eth2ClientTarget, EthClientRemote, ExecutionClient, InstalledPackageDet
 import * as db from "@dappnode/db";
 import { eventBus } from "@dappnode/eventbus";
 import { logs } from "@dappnode/logger";
-import { getConsensusUserSettings, getStakerConfigByNetwork } from "../stakerConfig/utils.js";
+import { getConsensusUserSettings } from "../stakerConfig/utils.js";
 import { packageGet } from "../../calls/packageGet.js";
 import { packageInstall } from "../../calls/packageInstall.js";
 import { packageRemove } from "../../calls/packageRemove.js";
@@ -360,14 +360,15 @@ export class EthereumClient {
     compose.write();
   }
 
-  private removeFullnodeAliasFromCompose<T extends Network>({
+  // TODO: Function should be private
+  public removeFullnodeAliasFromCompose<T extends Network>({
     execClientDnpName,
     execClientServiceName,
     fullnodeAlias = params.FULLNODE_ALIAS,
   }: {
     execClientDnpName: ExecutionClient<T>,
     execClientServiceName: string,
-    fullnodeAlias: string,
+    fullnodeAlias?: string,
   }): void {
     this.editFullnodeAliasInCompose({
       action: ComposeEditorAction.REMOVE,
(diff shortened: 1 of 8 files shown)


pablomendezroyo (approved):

packages/dappmanager/src/modules/ethClient/ethereumClient.ts
```diff
@@ -104,98 +110,158 @@ export class EthereumClient {
...
+  async updateFullnodeAlias<T extends Network>({
+    prevExecClientDnpName,
+    newExecClientDnpName,
+    network = "mainnet" as T
   }: {
-    dnpName: ExecutionClientMainnet;
-    removeAlias: boolean;
+    prevExecClientDnpName?: ExecutionClient<T>;
+    newExecClientDnpName?: ExecutionClient<T>;
+    network?: T;
```
Marketen, line 144: why is `network` optional and default to mainnet? wouldnt it be better to make it explicitly clear what network we want and make it a mandatory parameter?

packages/dappmanager/src/modules/ethClient/ethereumClient.ts
```diff
@@ -104,98 +110,158 @@ export class EthereumClient {
...
+
+  private getServiceNameFromPackage(pkg: InstalledPackageDetailData): string {
+    return pkg.manifest?.mainService || pkg.containers[0].serviceName;
+  }
+
+  private getContainerNameFromService(pkg: InstalledPackageDetailData, serviceName: string): string {
+    return pkg.containers.find(container => container.serviceName === serviceName)?.containerName || pkg.containers[0].containerName;
+  }
+
+  private async removeContainerAliasFromDockerNetwork({ containerName, fullnodeAlias }: { containerName: string, fullnodeAlias: string }): Promise<void> {
```
Marketen, line 196: we could rename the parameter "fullnodeAlias" to "aliasToRemove" or something like that, right? Makes it clearer

packages/dappmanager/src/modules/ethClient/ethereumClient.ts
```diff
@@ -104,98 +110,158 @@ export class EthereumClient {
...
       params.DNP_PRIVATE_NETWORK_NAME,
-      previousEthClientContainerName
+      containerName,
+      endpointConfig
     );
-    await dockerNetworkConnect(
+
+  }
+
+  private async addContainerAliasToDockerNetwork({ containerName, fullnodeAlias }: { containerName: string, fullnodeAlias: string }): Promise<void> {
```
Marketen, line 214: Same as above here, alias to add does not have to be always the fullnode alias itself, so we could rename "fullnodeAlias" to "aliasToAdd"

packages/dappmanager/src/modules/ethClient/ethereumClient.ts
dappnodedev: I chose to do it in this way because "mainnet" is the actual purpose of the fullnode functionality and before we did not consider fullnode domain in other networks so it should be the main one

dappnodedev:
As a side note, `<network>.fullnode.dappnode` domain will start working by the time `nsupdate` module is removed, i.e. when the PR https://github.com/dappnode/DNP_DAPPMANAGER/pull/1645 is merged

PR #1644, opened 2023-10-17
Fix db module not found

dappnodedev:
DB module was not added to `Dockerfile`, so the dappmanager did not start


pablomendezroyo:
Nice catch!

PR #1645, opened 2023-10-17
Remove `nsupdate` module

dappnodedev:
Overrides https://github.com/dappnode/DNP_DAPPMANAGER/pull/1588


packages/dappmanager/src/initializeDb.ts
```diff
@@ -211,6 +211,7 @@ export default async function initializeDb(): Promise<void> {
   // - Updates the domain: db.domain.set(domain);
   dyndns.generateKeys(); // Auto-checks if keys are already generated
 
+  // TODO: Move this functionality to Bind v2 package
```
dappnodedev, line 214: Here we are mapping the resolution of the `dyndns` to the private IP, instead of the public one, to improve UX. This should be done in the bind package instead. We can leave this here only until this is implemented

packages/dockerCompose/src/editor.ts
```diff
@@ -327,6 +327,9 @@ export class ComposeEditor {
         !Array.isArray(service.environment)
       )
         service.environment = stringifyEnvironment(service.environment);
+
+      // Remove hardcoded DNS
+      service.dns = undefined;
```
dappnodedev, line 332: We must check that this does not cause any trouble. Also, we should make sure the compose files of the packages are created without the `dns` instruction

packages/dockerCompose/test/unit/network.test.ts
```diff
@@ -55,6 +55,7 @@ describe("modules / compose / networks", () => {
             networks: {
               [networkName]: { aliases },
             },
+            dns: undefined,
```
dappnodedev, line 58: The deep equal test requires this to be added. Is it correct?

packages/dockerCompose/test/unit/network.test.ts
```diff
@@ -69,6 +70,7 @@ describe("modules / compose / networks", () => {
           "sample.dnp.dappnode.eth": {
             container_name,
             image,
+            dns: undefined,
```
dappnodedev, line 73: Same as above

pablomendezroyo (changes requested):
Unit test failing

dappnodedev:
Overwritten by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1748

PR #1646, opened 2023-10-18
Remove bind as DNS server from compose files

dappnodedev:
This is the first step to take in the dappmanager towards current Bind functionality removal


pablomendezroyo:
Is it documented somewhere the work of stopping from adding the bind to the compose file?

packages/dappmanager/src/modules/migrations/removeDnsFromComposeFiles.ts
```diff
@@ -0,0 +1,33 @@
...
+
+function removeDnsFromPackageComposeFile(dnpName: string, isCore: boolean): void {
+    const compose = new ComposeFileEditor(dnpName, isCore);
+    const services = compose.services();
+
+    for (const serviceName of Object.keys(services)) {
+        const composeService = services[serviceName].get();
+        if (composeService.dns) {
+            logs.info(`Removing DNS from ${serviceName} in ${dnpName} compose file`);
+            composeService.dns = undefined;
```
pablomendezroyo, line 29: Are you sure this instructions completely removes the `dns` field?

packages/dappmanager/test/integration/migrations/removeDnsFromCompose.test.int.ts
```diff
@@ -0,0 +1,113 @@
...
+            `${composeDnsRemovalTestPath}/test-dns-removal/docker-compose.yml`,
+            composeToBeMigratedBefore
+        );
+        // Compose already migrated
+        fs.writeFileSync(
+            `${composeDnsRemovalTestPath}/test-dns-removal/docker-compose-migrated.yml`,
+            composeAlreadyMigrated
+        );
+        // Redeclare global variables
+        params.DNCORE_DIR = composeDnsRemovalTestPath;
```
pablomendezroyo, line 65: You should not have to do this, the env `TEST=true` should change the dirs already allowing you to test propertly

pablomendezroyo (changes requested):

packages/dappmanager/test/integration/migrations/removeDnsFromCompose.test.int.ts
dappnodedev: I did not do this, it was already done in the compose migration test

packages/dappmanager/src/modules/migrations/removeDnsFromComposeFiles.ts
dappnodedev: Yes! This will remove the `dns` field from every `docker-compose.yml` file that still includes it

packages/dappmanager/test/integration/migrations/composeMigration.test.int.ts
```diff
@@ -2,103 +2,127 @@ import { expect } from "chai";
...
+    container_name: ${containerName}
+    restart: always
+    networks:
+      ${dncoreNetwork}:
+        ipv4_address: 172.33.1.7
+        aliases:
+          - ${serviceName}.test-migration.dappnode
+          - ${serviceName}.dappnode`;
+
   const composeAlreadyMigrated = `
```
Marketen, line 46: Now that this file has alias migration and dns migration tests, I would rename:
`composeAlreadyMigrated` to `composeAliasesToBeMigrated`
 `composeToBeMigratedBefore` to  `composeAliasesAlreadyMigrated`
 
 Or something like that, so we know that these are used for alias migration and not DNS

packages/dappmanager/test/integration/migrations/composeMigration.test.int.ts
```diff
@@ -2,103 +2,127 @@ import { expect } from "chai";
...
-      `${testMigrationPath}/test-migration/docker-compose.yml`,
+      `${testPackagePath}/docker-compose.yml`,
       { encoding: "utf8" }
     );
     expect(composeAfter.trim()).to.equal(composeAlreadyMigrated.trim());
-  });    
+  });
+
+  it("Should remove DNS from compose file", async () => {
+    removeDnsFromPackageComposeFile(container.dnpName, false);
```
Marketen, line 118: should we test also the `removeDnsFromPackageComposeFile` for core packages or is not worth it?

Marketen (approved):

pablomendezroyo (approved):

PR #1665, opened 2023-11-03
Improve Dockerfiles

dappnodedev:
Clean and improve production Dockerfile


pablomendezroyo (changes requested):
Could you do the same for the Dockerfile.dev

dappnodedev:
> Could you do the same for the Dockerfile.dev

Yes! It is still work in progress

PR #1667, opened 2023-11-06
fix get reboot required

pablomendezroyo:
fix get reboot required using a host script instead

### DIFF
--- a/packages/admin-ui/src/components/NotificationsMain.tsx
+++ b/packages/admin-ui/src/components/NotificationsMain.tsx
@@ -52,9 +52,8 @@ export default function NotificationsView() {
       id: "hostReboot",
       linkText: "Reboot",
       linkPath: systemPathName + "/" + systemSubPaths.power,
-      body:
-        "**Dappnode host reboot required.** Click **Reboot** to reboot the host and apply the changes",
-      active: rebootHostIsRequired
+      body: `**Dappnode host reboot required.** Click **Reboot** to reboot the host and apply the changes. The following packages will be updated: ${rebootHostIsRequired?.pkgs}`,
+      active: rebootHostIsRequired?.rebootRequired
     },
     /**
      * [SYSTEM-UPDATE]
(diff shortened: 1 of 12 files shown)


packages/admin-ui/src/components/NotificationsMain.tsx
```diff
@@ -52,9 +52,8 @@ export default function NotificationsView() {
       id: "hostReboot",
       linkText: "Reboot",
       linkPath: systemPathName + "/" + systemSubPaths.power,
-      body:
-        "**Dappnode host reboot required.** Click **Reboot** to reboot the host and apply the changes",
-      active: rebootHostIsRequired
+      body: `**Dappnode host reboot required.** Click **Reboot** to reboot the host and apply the changes. The following packages will be updated: ${rebootHostIsRequired?.pkgs}`,
```
dappnodedev, line 55: Will the user be redirected to the system page where they can trigger the reboot?

dappnodedev (approved):

packages/hostScripts/hostScripts/reboot_required.sh
```diff
@@ -0,0 +1,19 @@
...
+# Script to check weather a reboot is required or not.
+
+# Exit on error
+set -e
+
+if [ -f /var/run/reboot-required ]; then
+  # print reboot required and pkgs that require reboot
+  if [ -f /var/run/reboot-required.pkgs ]; then
+    # put pkgs in a string comma separated, if is just 1 pkg, do not add comma
+    pkgs=$(cat /var/run/reboot-required.pkgs | tr '\n' ',' | sed 's/,$//')
```
Marketen, line 12: are we 100% sure that all machines will have `tr` and `sed`?

pablomendezroyo: we cam assume that

packages/admin-ui/src/components/NotificationsMain.tsx
pablomendezroyo: yes!

PR #1670, opened 2023-11-07
Update and upgrade from UI

pablomendezroyo:
Add a button for updating and upgrading the host machine from UI


dappnodedev (changes requested):

packages/hostScripts/hostScripts/update_upgrade.sh
3alpha, line 1: This script should be run as a root so we shouldn't need `sudo` here.

3alpha (changes requested):

PR #1702, opened 2023-11-23
Added card to show the CPU Temperature

kdelap00:
Added a card in the UI that contains information about the temperature of the cpu.
Uses a bash script to get the temperature


packages/admin-ui/src/__mock-backend__/index.ts
```diff
@@ -372,7 +372,13 @@ export const otherCalls: Omit<Routes, keyof typeof namedSpacedCalls> = {
     }
   }),
   optimismConfigSet: async () => {},
-  updateUpgrade: async () => "Successfully updated"
+  updateUpgrade: async () => "Successfully updated",
+  
+ sensorsDataGet: async () => ({
```
dappnodedev, line 377: Identation here looks wrong. Install prettify extension in VS Code to fix it

packages/common/src/types.ts
```diff
@@ -944,6 +944,12 @@ export interface HostStatMemory {
   usedPercentage: number;
 }
 
+export type Sensors = {
+  temp1_input: number;
```
dappnodedev, line 948: I would follow the format of the rest of the types:
```
export type CpuTemperature = {
  currentTemperature: number;
  maxTemperature?: number;
  minTemperature?: number;
};
```

Note that I added the `?` because the value could be undefined, but adapt this code to your needs.

packages/dappmanager/src/calls/sensorsDataGet.ts
```diff
@@ -0,0 +1,9 @@
+import { Sensors } from "@dappnode/common";
+import { sensors } from "@dappnode/hostscriptsservices";
+
+/**
+ * Returns the sensors info of the host
+ */
+export async function sensorsDataGet(): Promise<Sensors> {
+    return await sensors();
```
dappnodedev, line 8: `sensors` is not verbose at all and functions/methods should include verbs, like `getCpuTemperature`

packages/hostScriptsServices/hostScripts/sensors.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
dappnodedev, line 1: `sensors.sh` is not a good name either, it needs to be more self-explanatory. What does the script do?

packages/hostScriptsServices/hostScripts/sensors.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Execute the first command
+output=$(paste <(cat /sys/class/thermal/thermal_zone*/type 2>/dev/null) <(cat /sys/class/thermal/thermal_zone*/temp 2>/dev/null) 2>/dev/null | column -s $'\t' -t | sed 's/\(.\)..$/.\1/')
```
dappnodedev, line 4: Same as above with `output`, it could be anything

packages/hostScriptsServices/hostScripts/sensors.sh
```diff
@@ -0,0 +1,12 @@
...
+echo "$output" | grep -E "x*_pkg_temp" | awk '{print $NF}'
```
dappnodedev, line 5: Shouldn't this be a `return` statement? Check other similar scripts in the repo

packages/hostScriptsServices/hostScripts/sensors.sh
dappnodedev, line 4: You should create variables to make the code cleaner. For example, you could create `THERMAL_ZONE_DIRS=/sys/class/thermal/thermal_zone*`

packages/hostScriptsServices/hostScripts/sensors.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Check if the output is empty or contains error message
+if [ -z "$output" ] || [[ "$output" == *"No such file or directory"* ]]; then
+    # If the first command fails, execute the second command
+    output=$(cat /sys/devices/pci0000:00/*/hwmon/hwmon*/temp1_input | sed 's/\(.\)..$/.\1/')
```
dappnodedev, line 10: This output and the one above are probably not following the same format, so you must convert it to a JSON of a pre-defined format that fits here

dappnodedev (changes requested):

packages/hostScriptsServices/hostScripts/sensors.sh
kdelap00: Other scripts use echo to return values.
Return statement can only return int values

packages/common/src/types.ts
kdelap00: Finally deleted this type.

packages/admin-ui/src/pages/dashboard/components/HostStats.tsx
```diff
@@ -18,22 +18,24 @@ const StatsCardContainer: React.FunctionComponent<{
...
+  const valueBar = Math.round(value);
 
   return (
     <>
       <ProgressBar
         variant={parseVariant(value)}
-        now={value}
-        label={value + "%"}
+        now={valueBar}
+        label={ valueType==="percentage" ? valueBar + "%"
```
pablomendezroyo, line 36: why not putting as argument directly the label?
```js
{ label: string}
```
and then calling the function with either the label "%" or "ºC", and if undefined take the `valueBar`

packages/hostScriptsServices/src/hostScripts/runScripts.ts
```diff
@@ -16,7 +16,8 @@ type ScriptName =
   | "collect_host_info.sh"
   | "avahi_daemon.sh"
   | "reboot_required.sh"
-  | "update_upgrade.sh";
+  | "update_upgrade.sh"
+  | "getCpuTemperature.sh";
```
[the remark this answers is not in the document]
pablomendezroyo: @kdelap00 ping

packages/hostScriptsServices/hostScripts/getCpuTemperature.sh
```diff
@@ -0,0 +1,14 @@
...
+THERMAL_ZONE_DIRS="/sys/class/thermal/thermal_zone*"
+
+# Execute the first command
+temp=$(paste <(cat $THERMAL_ZONE_DIRS/type 2>/dev/null) <(cat $THERMAL_ZONE_DIRS/temp 2>/dev/null) 2>/dev/null | column -s $'\t' -t | sed 's/\(.\)..$/.\1/')
+echo "$temp" | grep -E "x*_pkg_temp" | awk '{print $NF}'
+
+# Check if the output is empty or contains error message
+if [ -z "$temp" ] || [[ "$temp" == *"No such file or directory"* ]]; then
+    # If the first command fails, execute the second command
+    temp=$(cat /sys/devices/pci0000:00/*/hwmon/hwmon*/temp1_input | sed 's/\(.\)..$/.\1/')
```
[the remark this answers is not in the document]
pablomendezroyo: @kdelap00 ping

packages/hostScriptsServices/hostScripts/getCpuTemperature.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+THERMAL_ZONE_DIRS="/sys/class/thermal/thermal_zone*"
+
+# Execute the first command
+temp=$(paste <(cat $THERMAL_ZONE_DIRS/type 2>/dev/null) <(cat $THERMAL_ZONE_DIRS/temp 2>/dev/null) 2>/dev/null | column -s $'\t' -t | sed 's/\(.\)..$/.\1/')
+echo "$temp" | grep -E "x*_pkg_temp" | awk '{print $NF}'
+
+# Check if the output is empty or contains error message
+if [ -z "$temp" ] || [[ "$temp" == *"No such file or directory"* ]]; then
```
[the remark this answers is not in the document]
pablomendezroyo: @kdelap00 ping

packages/hostScriptsServices/hostScripts/getCpuTemperature.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+THERMAL_ZONE_DIRS="/sys/class/thermal/thermal_zone*"
+
+# Execute the first command
+temp=$(paste <(cat $THERMAL_ZONE_DIRS/type 2>/dev/null) <(cat $THERMAL_ZONE_DIRS/temp 2>/dev/null) 2>/dev/null | column -s $'\t' -t | sed 's/\(.\)..$/.\1/')
```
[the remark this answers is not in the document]
pablomendezroyo: @kdelap00 ping

packages/hostScriptsServices/hostScripts/getCpuTemperature.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
pablomendezroyo: @kdelap00 ping

packages/hostScriptsServices/src/hostScripts/scripts/getCpuTemperature.ts
```diff
@@ -0,0 +1,14 @@
+import { runScript } from "../runScripts.js";
+
+/**
```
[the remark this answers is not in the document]
pablomendezroyo: @kdelap00 ping

packages/dappmanager/src/calls/sensorsDataGet.ts
```diff
@@ -0,0 +1,8 @@
+import { getCpuTemperature } from "@dappnode/hostscriptsservices";
+
+/**
+ * Returns the sensors info of the host
+ */
+export async function sensorsDataGet(): Promise<number> {
```
pablomendezroyo, line 6: how confident we are about collecting the temperature from the host of the users? AFAIK we cannot ensure that every host machine from the user will have the paths/tools to collect the temperature, and in those cases, there will be shown an ugly error message.

Taking this into account I would consider this feature as optional since it could fail in some cases. So I would not throw the error if any, then the function could return `null` and the card will not be shown.

Something like 
```ts
/**
 * Returns the sensors info of the host
 */
export async function sensorsDataGet(): Promise<number | null> {
  try { 
  return await getCpuTemperature();
  } catch (e) {
  logs.warn(`Unable to get cpu temperature from the host: ${e}`)
  return null
  }
}

pablomendezroyo (changes requested):

packages/dappmanager/src/calls/sensorsDataGet.ts
kdelap00: Changed. Now if cannot get the temperature, it returns null. I also changed the frontend to not show the card if the function returns null

packages/hostScriptsServices/hostScripts/getCpuTemperature.sh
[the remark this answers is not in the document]
kdelap00: If paste command is not installed, the script will not return anything, but exit with code 0. So getCpuTemperature will return NaN. This output is not handled and the card will stay loading.

pablomendezroyo (approved):

pablomendezroyo:
Closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1833

PR #1733, opened 2023-12-11
Use remote gateway to render logos

pablomendezroyo:
Use remote gateway to render logos instead of local IPFS proxy

### DIFF
--- a/packages/utils/src/fileToGatewayUrl.ts
+++ b/packages/utils/src/fileToGatewayUrl.ts
@@ -1,5 +1,6 @@
 import { DistributedFile } from "@dappnode/common";
 import { normalizeHash } from "./normalizeHash.js";
+import url from "url";
 import { params } from "@dappnode/params";
 
 /**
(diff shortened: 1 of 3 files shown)


dappnodedev (approved):

packages/utils/src/fileToGatewayUrl.ts
```diff
@@ -13,7 +14,10 @@ export function fileToGatewayUrl(distributedFile?: DistributedFile): string {
 
   switch (distributedFile.source) {
     case "ipfs":
-      return `${params.IPFS_GATEWAY}${normalizeHash(distributedFile.hash)}`;
+      return url.resolve(
```
dappnodedev, line 17: What about:
```
return url.resolve(
  `${params.IPFS_REMOTE}/${params.IPFS_GATEWAY}`,
  normalizeHash(distributedFile.hash)
);
```
to avoid using `url.resolve` twice here?

PR #1744, opened 2023-12-12
Deprecate ethforward and bump ethers

pablomendezroyo:
Deprecate ethforward and bump ethers in dappmanager module


pablomendezroyo:
Will not implement

PR #1749, opened 2023-12-13
Fix first upload in frontend component `To.tsx`

pablomendezroyo:
Fix first upload in frontend component `To.tsx`, it only works the 2, 4 and so on


pablomendezroyo:
Does not fix it

PR #1755, opened 2023-12-18
Warning about OpenVPN deprecation

dappnodedev:
OpenVPN is about to be deprecated in favour of WireGuard. Here we show a warning for the users to migrate to the new VPN client
image


pablomendezroyo:
Not needed

PR #1765, opened 2023-12-26
Add service and timer to check docker network health

pablomendezroyo:
Add service and timer to check docker network health


packages/hostScriptsServices/hostScripts/check_docker_network.sh
```diff
@@ -0,0 +1,53 @@
...
+check_and_create_network() {
+    for ((i=0; i<max_retries; i++)); do
+        if docker network inspect "$network_name" &> /dev/null; then
+            echo "Docker network '$network_name' exists."
+            docker network inspect "$network_name"
+            return 0
+        else
+            echo "Attempt $((i+1)) of $max_retries: Docker network '$network_name' does not exist. Trying to create it..."
+
+            # [Insert your existing logic for checking core version and creating network here]
```
dappnodedev, line 18: This is incomplete. (Add TODO)

packages/hostScriptsServices/hostScripts/check_docker_network.sh
```diff
@@ -0,0 +1,53 @@
...
+        else
+            echo "Container $dappmanager_container is running. Status: $container_status"
+            return 0
+        fi
+    done
+    return 1
+}
+
+# Execute network check/create with retry logic
+check_and_create_network || echo "docker network $network_name has been recreated after $max_retries attempts."
```
dappnodedev, line 50: Are you sure the log will be printed when the functions returns 0?

packages/hostScriptsServices/src/hostTimers/runTimer.ts
```diff
@@ -0,0 +1,34 @@
...
+  const timerPath = path.resolve(params.HOST_TIMERS_SOURCE_DIR, timer);
+  try {
+    // Check if timer exists
+    if (!fs.existsSync(timerPath)) throw Error(`Host timer ${timer} not found`);
+
+    // Copy timer into shared volume
+    await copyHostTimer(timer);
+
+    // Enable timer
+    await shellHost(`systemctl enable ${timer}`);
```
dappnodedev, line 27: This function should not be callable from any API (I don't think it is, but just in case) or/and we should parse the input sent to the shell

dappnodedev (changes requested):

packages/hostScriptsServices/hostScripts/check_docker_network.sh
pablomendezroyo: Added!

packages/hostScriptsServices/hostScripts/check_docker_network.sh
pablomendezroyo: Verified!

packages/hostScriptsServices/src/hostTimers/runTimer.ts
pablomendezroyo: Nice catch! Ive just verified none api call it and I ve just removed the args so no inject scripting

packages/hostScriptsServices/hostScripts/check_docker_network.sh
```diff
@@ -0,0 +1,48 @@
...
+network_name="dncore_network"
+max_retries=5
+retry_delay=60
+
+# Function to check and create network with retry logic
+check_and_create_network() {
+    for ((i=0; i<max_retries; i++)); do
+        if docker network inspect "$network_name" &> /dev/null; then
+            echo "Docker network '$network_name' exists."
+            docker network inspect "$network_name"
```
dappnodedev, line 12: Why do you run it again? Do you want to actually print the output? In that case you might want to store the output and print it after, rather than executing again

packages/hostScriptsServices/hostScripts/check_docker_network.sh
```diff
@@ -0,0 +1,48 @@
...
+            return 0
+        else
+            echo "ATTENTION! Docker network '$network_name' does not exist. Creating it..."
+            # Check the core docker image tag to get the version and determine the subnet to be used for creating the docker core network
+            core_version="$(docker image ls --filter reference="core.dnp.dappnode.eth" --format '{{.Tag}}' | sort -V | tail -n 1)"
+            echo "core version retrieved $core_version"
+
+            # Compare with 0.2.30 using sort
+            if printf '0.2.30\n%s\n' "$core_version" | sort -V | head -n 1 | grep -q '0.2.30'; then
```
dappnodedev, line 21: New version of the Core with the private IP range will be `v0.3.0` !!

dappnodedev: Remember to change all appearances in comments and logs

packages/hostScriptsServices/hostScripts/check_docker_network.sh
```diff
@@ -0,0 +1,48 @@
...
+                # core_version is greater than or equal to 0.2.30
+                echo "Core version is greater than or equal to 0.2.30, using subnet $subnet"
+            else
+                subnet="10.20.0.0/24"
+                # core_version is less than 0.2.30
+                echo "Core version is less than 0.2.30, using subnet $subnet"
+            fi
+
+            # create docker network
+            docker network create --driver bridge --subnet "$subnet" "$network_name"
```
dappnodedev, line 32: If this fails we might want to create the network without specifying the subnet so that we are sure it is created and then the dappmanager will be responsible for migrating it to the proper subnet

dappnodedev (changes requested):

packages/hostScriptsServices/hostScripts/check_docker_network.sh
pablomendezroyo: It might be usefull to know the docker network details, but will remove it

packages/hostScriptsServices/hostScripts/check_docker_network.sh
pablomendezroyo: true! nice catch

packages/hostScriptsServices/hostScripts/check_docker_network.sh
pablomendezroyo: implemented!

PR #1772, opened 2023-12-28
Remove connected containers if disconnect fails in IP range migration

pablomendezroyo:
Remove connected containers if disconnect fails in IP range migration


packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
```diff
@@ -0,0 +1,87 @@
...
+      logs.warn(`docker network ${networkName} not found, creating it...`);
+
+      await removeNetworksOverlappingSubnetIfNeeded(networkSubnet).catch((e) =>
+        logs.error(`error removing overlapping networks: ${e}`)
+      );
+      // CRITICAL: if this step fails migration failure
+      const network = await docker.createNetwork(networkOptions);
+      return { network, isNetworkRecreated: true };
+    } else {
+      // TODO: What do we do here?
```
dappnodedev, line 83: We need to handle this. We could try to find out all the errors that can be thrown and act accordingly

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
```diff
@@ -0,0 +1,87 @@
...
+        networkOptions
+      );
+      return { network, isNetworkRecreated: true };
+    }
+  } catch (e) {
+    if (e.statusCode === 404) {
+      // docker network not found, create it
+      logs.warn(`docker network ${networkName} not found, creating it...`);
+
+      await removeNetworksOverlappingSubnetIfNeeded(networkSubnet).catch((e) =>
```
dappnodedev, line 76: We cannot omit this error, it will not let us create the network

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/recreateDockerNetwork.ts
```diff
@@ -0,0 +1,71 @@
...
+  newNetworkOptions: Dockerode.NetworkCreateOptions
+): Promise<Dockerode.Network> {
+  logs.info(`disconnecting all containers from ${networkToRemove.id}`);
+  await disconnectAllContainersFromNetwork(networkToRemove).catch((e) =>
+    logs.error(
+      `error while disconnecting all containers from the network: ${e.message}`
+    )
+  );
+
+  await removeConnectedContainersIfAny(networkToRemove).catch((e) =>
```
dappnodedev, line 30: I would retry to disconnect the containers before removing them

dappnodedev (changes requested):

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
pablomendezroyo: Is not part of this PR

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
pablomendezroyo: But it worths trying to continue with the flow instead of blocking it.

Also that function does `listNetworks()` which could fail without meaning there are overlapping networks

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/recreateDockerNetwork.ts
pablomendezroyo: this is actually the flow, isnt it?

dappnodedev: I don't think so. We try to disconnect them and then we directly remove them. Removing those containers should be the last resort as the container might be the `web3signer` for example. I would first retry disconnecting the containers for some attempts

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
dappnodedev: But we should try to "fix" the errors that might be thrown there. We might need to find any edge cases that could happen there and act accordingly

pablomendezroyo: As I said the `docker.listNetworks()` might fail, and that would prevent the flow of creating the network while it could be allowed to be created.

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/recreateDockerNetwork.ts
pablomendezroyo: So it is, the flow:
1. Disconnect them
2. If still some docker containers connected, then remove them

As you said we could do a removal retry if there are still some connected containers but, how many times, during how much time?

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
dappnodedev: It could fail, but I'd say it is extremely weird that listing the networks fails.... And that would not be a reason why not to try to handle the errors that might be thrown (or just retry), even if you end up trying to create the network anyway

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/recreateDockerNetwork.ts
dappnodedev: I would retry like 5 times or so. Leaving a delay of some seconds

dappnodedev: Do we know what happens if we try to disconnect/connect a restarting container from the network?

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/recreateContainersToRecreate.ts
```diff
@@ -0,0 +1,32 @@
...
+): Promise<void> {
+  if (containersToRecreate.length > 0) {
+    logs.info(
+      `Recreating docker containers that require to be recreated: ${containersToRecreate}`
+    );
+    const composeFilesPathsToRecreate = (
+      await Promise.all(
+        filterContainers(containersToRecreate).map(async (cn) => {
+          // get the compose file path
+          return (await docker.getContainer(cn).inspect()).Config.Labels[
```
dappnodedev, line 16: This could be another function called `getComposeFileFromContainer`? Also, we must be sure that the file path of the docker compose is the one inside the dappmanager (and not the one in the host). I am pretty sure this code is ok, but just in case

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/restartContainersToRestart.ts
```diff
@@ -0,0 +1,19 @@
+import { docker } from "@dappnode/dockerapi";
+import { logs } from "@dappnode/logger";
+import { filterContainers } from "./filterContainers.js";
+
+export async function restartContainersToRestart(
```
dappnodedev, line 5: `restartForceStoppedContainers`maybe?

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/restoreContainersToNetwork.ts
```diff
@@ -0,0 +1,31 @@
+import Dockerode from "dockerode";
+import { reconnectDisconnectedContainers } from "./reconnectDisconnectedContainers.js";
+import { recreateContainersToRecreate } from "./recreateContainersToRecreate.js";
+import { restartContainersToRestart } from "./restartContainersToRestart.js";
+
+export async function restoreContainersToNetwork({
```
dappnodedev, line 6: I think all the functions called inside this one should have a `catch` that makes sure that an error thrown does not prevent the others from being executed

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/restartContainersToRestart.ts
```diff
@@ -0,0 +1,19 @@
...
+  containersToRestart: string[]
+): Promise<void> {
+  if (containersToRestart.length > 0) {
+    logs.info(
+      `Restarting docker containers that require to be restarted: ${containersToRestart}`
+    );
+
+    await Promise.all(
+      filterContainers(containersToRestart).map(async (cn) => {
+        await docker.getContainer(cn).restart();
```
dappnodedev, line 15: Can an error thrown in one of the promises make the rest fail?

dappnodedev (changes requested):

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/ensureDockerNetworkConfig.ts
pablomendezroyo: still, it worth moving forward. Ofcourse, taking into account as many edge cases as possible

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/removeNetworksOverlappingSubnetIfNeeded.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (overlappingNetworks.length > 0) {
+    logs.info(
+      `Found ${overlappingNetworks.length} networks to remove (overlapping subnet)`
+    );
+
+    await Promise.all(
+      overlappingNetworks.map(async (networkInfo) => {
+        const networkName = networkInfo.Name;
+        const network = docker.getNetwork(networkName);
```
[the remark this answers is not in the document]
pablomendezroyo: done!

packages/migrations/src/migrateDockerNetworkIpRange/ensureDockerNetworkConfig/removeNetworksOverlappingSubnetIfNeeded.ts
```diff
@@ -0,0 +1,36 @@
...
+  const overlappingNetworks = networks.filter((network) =>
+    isNetworkOverlappingSubnet(network, networkSubnet)
+  );
+
+  if (overlappingNetworks.length > 0) {
+    logs.info(
+      `Found ${overlappingNetworks.length} networks to remove (overlapping subnet)`
+    );
+
+    await Promise.all(
```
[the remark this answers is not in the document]
pablomendezroyo: The rest of the promises will be executed although the call `Promise.all` will reject

packages/migrations/src/migrateDockerNetworkIpRange/migrateDockerNetworkIpRange.ts
```diff
@@ -31,10 +32,19 @@ export async function migrateDockerNetworkIpRange({
   };
 }): Promise<void> {
   const aliasesMap = await getNetworkAliasesMapNotThrow(dockerNetworkName);
-
-  const { network, isNetworkRecreated } = await ensureDockerNetworkConfig({
+  const networkContainersNamesAndIps = await getNetworkContainerNamesAndIps(
```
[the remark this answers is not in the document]
pablomendezroyo: merged!

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/filterContainers.ts
```diff
@@ -0,0 +1,8 @@
+import { params } from "@dappnode/params";
+
+export function filterContainers(containers: string[]): string[] {
```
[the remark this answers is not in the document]
pablomendezroyo: renamed!

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/recreateContainersToRecreate.ts
```diff
@@ -0,0 +1,32 @@
+import { docker, dockerComposeUp } from "@dappnode/dockerapi";
+import { logs } from "@dappnode/logger";
+import { filterContainers } from "./filterContainers.js";
+
+export async function recreateContainersToRecreate(
```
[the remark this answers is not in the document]
pablomendezroyo: agree!

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/restartContainersToRestart.ts
pablomendezroyo: agree!

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/restartContainersToRestart.ts
pablomendezroyo: No!

packages/migrations/src/migrateDockerNetworkIpRange/restoreContainersToNetwork/restoreContainersToNetwork.ts
pablomendezroyo: done!

PR #1785, opened 2024-01-12
bump version number to current for testing pre release

alexpeterson91:
bumping version number to the current release since the hardcoded version is 8 versions behind current, messing up communication between some packages i.e. core and packages requiring minimum dappmanger versions,  

This is only for testing purposes as when its merged and published it will be bumped to v0.2.80

## Context

## Approach


alexpeterson91:
resolved in #1782

PR #1804, opened 2024-01-23
- Add Dappnode Peer to Local Ipfs Node

mateumiralles:
Add Dappnode Peer to Local Ipfs Node

### DIFF
--- a/packages/migrations/src/addDappnodePeerToLocalIpfsNode.ts
+++ b/packages/migrations/src/addDappnodePeerToLocalIpfsNode.ts
@@ -1,13 +1,14 @@
 import { logs } from "@dappnode/logger";
 
+//Adds Dappnode IPFS peer to the local IPFS node for more stability when fetching IPFS content
 export async function addDappnodePeerToLocalIpfsNode() {
-  const IpfsPeer =
+  const dappnodeIpfsPeer =
     "/ip4/167.86.114.131/tcp/4001/p2p/QmfB6dT5zxUq1BXiXisgcZKYkvjywdDYBK5keRaqDKH633";
-  const IpfsAlias = "ipfs.dappnode";
+  const ipfsAlias = "ipfs.dappnode";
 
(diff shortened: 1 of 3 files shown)


packages/migrations/src/addDappnodePeerToLocalIpfsNode.ts
```diff
@@ -0,0 +1,15 @@
+import { logs } from "@dappnode/logger";
+
+export async function addDappnodePeerToLocalIpfsNode() {
+  const IpfsPeer =
```
pablomendezroyo, line 4: follow js standards for variables declarations. Start with non-capital letter

pablomendezroyo: Also consider naming it `dappnodeIpfsPeer`

packages/migrations/src/addDappnodePeerToLocalIpfsNode.ts
```diff
@@ -0,0 +1,15 @@
+import { logs } from "@dappnode/logger";
+
+export async function addDappnodePeerToLocalIpfsNode() {
```
pablomendezroyo, line 3: I would add a comment to the function: what it does, and why is it necessary

packages/migrations/src/addDappnodePeerToLocalIpfsNode.ts
```diff
@@ -0,0 +1,15 @@
...
+  const IpfsPeer =
+    "/ip4/167.86.114.131/tcp/4001/p2p/QmfB6dT5zxUq1BXiXisgcZKYkvjywdDYBK5keRaqDKH633";
+  const IpfsAlias = "ipfs.dappnode";
```
pablomendezroyo, line 6: same as above

pablomendezroyo (changes requested):

pablomendezroyo (approved):

PR #1812, opened 2024-01-25
Add recreate dappnode to latest service

pablomendezroyo:
Add recreate dappnode to latest service

### DIFF
--- a/packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
+++ b/packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
@@ -5,10 +5,10 @@
 # This script is a debian host service that ensures the minimum requirement for dappnode to be healthy: dappmanager docker container is running
 # - If dappmanager docker compose file does not exist, then recreate DAppNode to latest
 # - If dappmanager docker container does not exist, then run docker compose up to dappmanager file
-# - If dappmanager docker container is not running, then restart it
-# - If dappmanager docker container is restarting, then check 5 times with a 5 seconds delay the status is still the same. If 3 or more times the status is the same, then recreate dappnode
-# - If dappmanager docker container is created, exited, dead or paused, then restart it
-# - If dappmanager docker container is running, then do nothing
+# - If dappmanager docker container is restarting for more than 25 seconds, then recreate DAppNode to latest
+# - If dappmanager docker container is created or exited or dead or paused, then restart dappmanager docker container
+#   - If previous restart fails, then run docker compose up to dappmanager file
+#       - If previous docker compose up fails, then recreate DAppNode to latest
 
 DNCORE_DIR="/usr/src/dappnode/DNCORE"
 DAPPMANAGER_DNCORE_FILE="$DNCORE_DIR/docker-compose-dappmanager.yml"
@@ -29,7 +29,7 @@ if [ ! -f "$DAPPMANAGER_DNCORE_FILE" ]; then
 fi 
 
 # dappmanager docker container does not exist, then run docker compose up to dappmanager file
-if [ ! "$(docker ps -q -f -a name="$DAPPMANAGER_CONTAINER_NAME")" ]; then
+if [ ! "$(docker ps -aq -f name="$DAPPMANAGER_CONTAINER_NAME")" ]; then
     echo "Container $DAPPMANAGER_CONTAINER_NAME does not exist. Running docker compose up to dappmanager file." | tee "$RECREATE_DAPPNODE_LOG_FILE"
     docker-compose -f "$DAPPMANAGER_DNCORE_FILE" up -d 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
     # if the previous command fails, then recreate dappnode
@@ -71,8 +71,14 @@ elif [ "$STATUS" == "created" ] || [ "$STATUS" == "exited" ] || [ "$STATUS" == "
     docker restart "$DAPPMANAGER_CONTAINER_NAME" 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
     # if the previous command fails, then recreate dappnode
     if [ $? -ne 0 ]; then
-        echo "Error while docker restart to $DAPPMANAGER_CONTAINER_NAME. Recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
-        wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        # do docker compose up
+        echo "Error while docker restart to $DAPPMANAGER_CONTAINER_NAME. Running docker compose up to dappmanager file." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        docker-compose -f "$DAPPMANAGER_DNCORE_FILE" up -d 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        # if the previous command fails, then recreate dappnode
+        if [ $? -ne 0 ]; then
+            echo "Error while docker compose up $DAPPMANAGER_DNCORE_FILE. Recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+            wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        fi
     fi
 else
     # do nothing


packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# Host script service executed with a timer to ensure dappmanager container is running, otherwise recreate dappnode to latest version
+# The edge case where the dappmanager container does not exist is not covered
```
dappnodedev, line 4: We could easily handle it, right? We could do `docker compose up` of the Dappmanager in DNCORE directory

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Name of the dappmanager Docker container
+DAPPMANAGER_CONTAINER_NAME="DAppNodeCore-dappmanager.dnp.dappnode.eth"
+
+# Check the status of the container
+# Possible docker container status: created, exited, dead, created, restarting, paused and running
+STATUS=$(docker inspect --format='{{.State.Status}}' "$DAPPMANAGER_CONTAINER_NAME" 2>/dev/null)
+
+# For safety check for docker container status not expected instead of checking for running
```
dappnodedev, line 13: Why wouldn't `!= running` work? I do not see a case where this makes sense, except for the one in which you have different behavior for each of them

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,38 @@
...
+if [ "$STATUS" == "created" ] || [ "$STATUS" == "exited" ] || [ "$STATUS" == "dead" ] || [ "$STATUS" == "created" ] || [ "$STATUS" == "restarting" ] || [ "$STATUS" == "paused" ]; then
```
dappnodedev, line 14: "created" is duplicated

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,38 @@
...
+        # check the status of the container
+        STATUS=$(docker inspect --format='{{.State.Status}}' "$DAPPMANAGER_CONTAINER_NAME" 2>/dev/null)
+        # if the status is the same as before, then increase the counter
+        if [ "$STATUS" == "created" ] || [ "$STATUS" == "exited" ] || [ "$STATUS" == "dead" ] || [ "$STATUS" == "created" ] || [ "$STATUS" == "restarting" ] || [ "$STATUS" == "paused" ]; then
+            ((counter++))
+        fi
+        sleep 5
+    done
+    # if the counter is 3 or more, then recreate dappnode
+    if [ "$counter" -ge 3 ]; then
```
dappnodedev, line 29: I would try other things before reinstalling:
1. If the container is not in an unrecoverable status (restarting for example) I would try to restart it
2. If that does not work I'd try to do `docker compose up` of the dappmanager compose file (maybe checking that the dappmanager image defined in the compose exists locally)
3. If 2 and 3 do not work, then reinstall

dappnodedev (changes requested):

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
pablomendezroyo: You are right that would work and would be perfectly fine, but I am afraid of the behaviour of this script, specially in unexpected bug loops. I rather prefer to check for those positives instead of negatives.

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
pablomendezroyo: Yeah, will add it

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
pablomendezroyo: done

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# dappmanager docker compose file does not exist, then install DAppNode
+if [ ! -f "$DAPPMANAGER_DNCORE_FILE" ]; then
+    echo "File $DAPPMANAGER_DNCORE_FILE does not exist. Installing DAppNode." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE" 
+    return
+fi 
+
+# dappmanager docker container does not exist, then run docker compose up to dappmanager file
+if [ ! "$(docker ps -q -f name="$DAPPMANAGER_CONTAINER_NAME")" ]; then
```
dappnodedev, line 22: Don't you need to add flag `-a` here? If the dappmanager is stopped we might just need to restart it, not to do `compose up`

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,70 @@
...
+            ((counter++))
+        fi
+        sleep 5
+    done
+    # if the counter is 3 or more, then recreate dappnode
+    if [ "$counter" -ge 3 ]; then
+        echo "Container $DAPPMANAGER_CONTAINER_NAME is restarting after 25s. Status: $STATUS. Recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    else
+        echo "Container $DAPPMANAGER_CONTAINER_NAME is not running. Status: $STATUS. Double checking before recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
```
dappnodedev, line 57: This log looks wrong

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,70 @@
...
+        # if the status is the same as before, then increase the counter
+        if [ "$STATUS" == "restarting" ]; then
+            ((counter++))
+        fi
+        sleep 5
+    done
+    # if the counter is 3 or more, then recreate dappnode
+    if [ "$counter" -ge 3 ]; then
+        echo "Container $DAPPMANAGER_CONTAINER_NAME is restarting after 25s. Status: $STATUS. Recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
```
dappnodedev, line 55: Why don't you try here doing `compose up` before reinstalling?

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,70 @@
...
+    else
+        echo "Container $DAPPMANAGER_CONTAINER_NAME is not running. Status: $STATUS. Double checking before recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    fi
+elif [ "$STATUS" == "created" ] || [ "$STATUS" == "exited" ] || [ "$STATUS" == "dead" ] || [ "$STATUS" == "paused" ]; then
+    echo "Container $DAPPMANAGER_CONTAINER_NAME is not running. Status: $STATUS. Running docker compose up to dappmanager file." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    docker-compose -f "$DAPPMANAGER_DNCORE_FILE" up -d 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
```
dappnodedev, line 61: Why don't you try here doing `docker restart` before `compose up`?

dappnodedev (changes requested):

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
pablomendezroyo: nice catch!

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,80 @@
...
+
+# dappmanager docker compose file does not exist, then install DAppNode
+if [ ! -f "$DAPPMANAGER_DNCORE_FILE" ]; then
+    echo "File $DAPPMANAGER_DNCORE_FILE does not exist. Installing DAppNode." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE" 
+    return
+fi 
+
+# dappmanager docker container does not exist, then run docker compose up to dappmanager file
+if [ ! "$(docker ps -q -f -a name="$DAPPMANAGER_CONTAINER_NAME")" ]; then
```
dappnodedev, line 32: This would be: 
`docker ps -aq -f name="$DAPPMANAGER_CONTAINER_NAME"`

dappnodedev: `-f` must be place just before the filter. It will not work as it is now

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
dappnodedev: I would use reinstalling as the last option. If there is a slight chance that doing compose up works I think it is worth trying before reinstalling

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
```diff
@@ -0,0 +1,80 @@
...
+        echo "Container $DAPPMANAGER_CONTAINER_NAME is restarting after 25s. Status: $STATUS. Recreating dappnode to latest version." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+        wget -O - https://installer.dappnode.io | sudo UPDATE=true bash 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    else
+        echo "Container $DAPPMANAGER_CONTAINER_NAME recovered from restarting state. Doing nothing." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    fi
+elif [ "$STATUS" == "created" ] || [ "$STATUS" == "exited" ] || [ "$STATUS" == "dead" ] || [ "$STATUS" == "paused" ]; then
+    echo "Container $DAPPMANAGER_CONTAINER_NAME is not running. Status: $STATUS. Running docker restart to dappmanager." | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    docker restart "$DAPPMANAGER_CONTAINER_NAME" 2>&1 | tee "$RECREATE_DAPPNODE_LOG_FILE"
+    # if the previous command fails, then recreate dappnode
+    if [ $? -ne 0 ]; then
```
dappnodedev, line 73: Same here, I would first try compose up too before reinstalling

packages/hostScriptsServices/hostScripts/recreate_dappnode.sh
dappnodedev: Reinstalling means that you need to rely on a third party server and you also need to have Internet connection. However, doing compose up will work anyway if the image is locally found. Then if dappmanager is still not active, we can go for reinstallation

dappnodedev (changes requested):

PR #1815, opened 2024-01-26
Move utils required by sdk to types

pablomendezroyo:
Move utils required by the dappnode sdk to types


packages/types/src/utils/getContainerName.ts
```diff
@@ -19,6 +18,6 @@ export const getContainerName = ({
 }): string =>
   // Note: _PREFIX variables already end with the character "-"
   [
-    isCore ? params.CONTAINER_CORE_NAME_PREFIX : params.CONTAINER_NAME_PREFIX,
+    isCore ? "DAppNodeCore-" : "DAppNodePackage-",
```
dappnodedev, line 21: Why hardcoding it? I'd keep the variables

pablomendezroyo: Those values will never change, and this way I avoid having to import the params module which otherwise will make the npm package erroring cause the params module only exists in local

Issue #1810, opened 2024-01-24
zkEVM rollups


GiselleNessi:
Integration of ZKEVM in the Rollups new tab

In the "Rollups > ZKEVM" section, we need:

- A single card for the package
- Know whether the package is installed or not
- If the package is installed, show it as selected if it is running and unselected if not
- If the package is not installed, show it as unselected
- If the user unselects the package and hits "apply", the package should be stopped
- If the user selects the package and hits "apply" the package should be (installed) + started
- If the package is installed and running, a button should be shown inside the card ("Go to UI")
- To install the package you need to know its identifier, which is its dnpName. In this case, the dnpName of the package will be: zkevm-tokens-withdrawal.dnp.dappnode.eth

wip
<img width="1304" alt="Screenshot 2024-01-24 at 9 47 21 PM" src="https://github.com/dappnode/DNP_DAPPMANAGER/assets/53020124/ac21f5c6-34b6-40f0-ad43-bd9b5a9061a3">

PR #1838, opened 2024-02-12
zkevm rollup new tab

GiselleNessi:
Rollups > ZKEVM

[Link to issue ](https://github.com/dappnode/DNP_DAPPMANAGER/issues/1810)

Know whether the package is installed or not
If the package is installed, show it as selected if it is running and unselected if not
If the package is not installed, show it as unselected
If the user unselects the package and hits "apply", the package should be stopped
If the user selects the package and hits "apply" the package should be (installed) + started
If the package is installed and running, a button should be shown inside the card ("Go to UI")
To install the package you need to know its identifier, which is its dnpName. In this case, the dnpName of the package will be: zkevm-tokens-withdrawal.dnp.dappnode.eth

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,9 @@
     "clean:libraries": "rimraf packages/**/node_modules && rimraf node_modules",
     "clean:build": "rimraf packages/**/dist && rimraf packages/**/build"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@dappnode/types": "0.1.35"
+  },
   "devDependencies": {
     "@types/chai": "^4.3.11",
     "@types/mocha": "^10.0.6",
--- a/packages/admin-ui/src/pages/rollups/components/Zkevm.tsx
+++ b/packages/admin-ui/src/pages/rollups/components/Zkevm.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { ZKEVMItem } from "../../../../../types/src/rollups";
 import { useApi } from "api";
 import { AppContext } from "App";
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -25,7 +25,9 @@
     "clean:libraries": "rimraf packages/**/node_modules && rimraf node_modules",
     "clean:build": "rimraf packages/**/dist && rimraf packages/**/build"
   },
-  "dependencies": {},
+  "dependencies": {
+    "@dappnode/types": "0.1.35"
```
pablomendezroyo, line 29: you should not install this dependendcy at the root level but in the paxkage.json inside the module that requires it

packages/admin-ui/src/pages/rollups/components/useZkEvmConfig.ts
```diff
@@ -0,0 +1,31 @@
+import { useState, useEffect } from "react";
+import { ZKEVMItem } from "../../../../../types/src/rollups";
```
pablomendezroyo, line 2: this item outside the module, you should import it as follows `from "@dappnode/types"`

pablomendezroyo (changes requested):

packages/admin-ui/src/pages/rollups/components/Zkevm.tsx
```diff
@@ -0,0 +1,126 @@
...
+              </Button>
+            </div>
+
+            {/* Render alert if changes are not allowed */}
+            {!isZkEvmInstalled && (
+              <>
+                <br />
+                <br />
+                <Alert variant="danger">
+                  Cannot apply changes: zkEVM is not installed
```
Marketen, line 106: When we cant apply changes, normally it is because no changes are detected.

I think its best to disable it by default and use booleans to define if it is enabled. See
https://github.com/dappnode/DNP_DAPPMANAGER/blob/gi/rollup-zkevm/packages/admin-ui/src/pages/stakers/components/StakerNetwork.tsx#L265

GiselleNessi:
opening a new pr for this issue

PR #1854, opened 2024-02-23
Notifying if internet disconnected with an alert

mateumiralles:
Warning to the user with an alert if the dappnode host is not connected to internet. Also adding a link to our documentation in Autodiagnose for this case.

### DIFF
--- a/packages/admin-ui/src/components/NotificationsMain.tsx
+++ b/packages/admin-ui/src/components/NotificationsMain.tsx
@@ -12,7 +12,8 @@ import {
 import {
   getWifiStatus,
   getPasswordIsSecure,
-  getRebootIsRequired
+  getRebootIsRequired,
+  getIsConnectedToInternet
 } from "services/dappnodeStatus/selectors";
 import {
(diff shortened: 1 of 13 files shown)


packages/admin-ui/src/components/NotificationsMain.tsx
```diff
@@ -44,6 +46,17 @@ export default function NotificationsView() {
...
   const notifications = [
+    /**
+     * [HOST-CONNECTED-TO-INTERNET]
+     * Tell the user if is connected to internet
+     */
+    {
+      id: "connectedToInternet",
+      linkText: "Check",
+      linkPath: "support/auto-diagnose",
+      body: `**Dappnode host is not connected to internet.** Click **Check** to run the autodiagnose and learn more about it.`,
```
pablomendezroyo, line 57: I would change the copy from "check" to "navigate" to autodiagnose to check the dappnode health.

pablomendezroyo (changes requested):

pablomendezroyo (approved):

PR #1866, opened 2024-03-04
Add minimumDockerVersion as optional requirement in manifest

mateumiralles:
When trying to install a package whose ```minimumDockerVersion``` in manifest is higher than the current docker version, the installation will be blocked and an alert with a btn will redirect to "Docker update" in "System / advanced"

### DIFF
--- a/packages/dappmanager/src/calls/fetchDnpRequest.ts
+++ b/packages/dappmanager/src/calls/fetchDnpRequest.ts
@@ -151,8 +151,7 @@ function getRequiresCoreUpdate(
     ? manifest.requirements.minimumDappnodeVersion
     : "";
   return Boolean(
-    manifest.requirements &&
-      minDnVersion &&
+    minDnVersion &&
       valid(minDnVersion) &&
       valid(coreVersion) &&
       gt(minDnVersion, coreVersion)
@@ -167,8 +166,7 @@ async function getRequiresDockerUpdate({
   if (!minDockerVersion) return false;
   const currentDockerVersion = await getDockerVersion();
   return Boolean(
-    manifest.requirements &&
-      minDockerVersion &&
+    minDockerVersion &&
       valid(minDockerVersion) &&
       valid(currentDockerVersion) &&
       gt(minDockerVersion, currentDockerVersion)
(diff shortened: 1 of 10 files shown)


packages/dappmanager/src/calls/fetchDnpRequest.ts
```diff
@@ -147,11 +152,28 @@ function getRequiresCoreUpdate(
...
+}): Promise<boolean> {
+  const minDockerVersion = manifest.requirements?.minimumDockerVersion;
+  if (!minDockerVersion) return false;
+  const currentDockerVersion = await getDockerVersion();
+  return Boolean(
+    manifest.requirements &&
+      minDockerVersion &&
+      valid(minDockerVersion) &&
+      valid(currentDockerVersion) &&
+      gt(minDockerVersion, currentDockerVersion)
```
pablomendezroyo, line 174: shouldn't be used greater OR equal than instead of just greater than?

packages/dappmanager/src/calls/fetchDnpRequest.ts
```diff
@@ -147,11 +152,28 @@ function getRequiresCoreUpdate(
...
+  manifest
+}: {
+  manifest: Manifest;
+}): Promise<boolean> {
+  const minDockerVersion = manifest.requirements?.minimumDockerVersion;
+  if (!minDockerVersion) return false;
+  const currentDockerVersion = await getDockerVersion();
+  return Boolean(
+    manifest.requirements &&
+      minDockerVersion &&
```
pablomendezroyo, line 171: the `manifest.requirements &&` can be removed since `minDockerVersion` already checks that

pablomendezroyo (changes requested):

packages/dappmanager/src/calls/fetchDnpRequest.ts
pablomendezroyo: Does not make sense

packages/dappmanager/src/calls/fetchDnpRequest.ts
mateumiralles: done

pablomendezroyo (approved):

PR #1870, opened 2024-03-06
Fix packageInstall/remove when https package was stopped

pablomendezroyo:
Move `getIsRunningHttps` to class  instead of calling this utility function each time there was a call to the https portal.

Also join utilities functions into https portal class


packages/httpsPortal/src/httpsPortal.ts
```diff
@@ -88,6 +95,7 @@ export class HttpsPortal {
    * Remove an internal container from being exposed to the external internet
    */
   async removeMapping(mapping: HttpsPortalMapping): Promise<void> {
+    if (!(await this.isRunningHttps())) return;
```
dappnodedev, line 98: Shouldn't we throw an error here?

pablomendezroyo: probably although it would require a deeper research, I will open an issue so we can push this for this release

PR #1876, opened 2024-03-11
Ethical metrics - add telegram implementation into backend

pablomendezroyo:
Implement telegram channel id in the backend:

- Change the ethical metrics settings in the database
- Add migration to database
- Change type
- Add implementation of tgChannelId in ethical metrics register


packages/admin-ui/src/components/welcome/features/EnableEthicalMetrics.tsx
```diff
@@ -30,9 +30,11 @@ export default function EnableEthicalMetrics({
 
   function onSetEnanleEthicalMetrics() {
```
mateumiralles, line 31: change misspelling to "Enable"

mateumiralles (changes requested):

Issue #1865, opened 2024-03-04
Add MEV-Boost package to Stakers Holesky


Marketen:
MEV-Boost now supports holesky testnet: https://github.com/flashbots/mev-boost?tab=readme-ov-file#holesky-testnet
It needs to be added in Holesky stakers page. Prater is close to being deprecated and its relays are not maintained anymore.

MEV-Boost holesky should have the same UX as MEV-Boost in other networks:
![Screenshot_20240304_110426](https://github.com/dappnode/DNP_DAPPMANAGER/assets/36164126/b3b5f699-7b80-4d86-8147-01605a373296)

Working relays to add in MEV-Boost holesky:
Flashbots: https://0xafa4c6985aa049fb79dd37010438cfebeb0f2bd42b115b89dd678dab0670c1de38da0c4e9138c9290a398ecd9a0b3110@boost-relay-holesky.flashbots.net
Aestus: https://0xab78bf8c781c58078c3beb5710c57940874dd96aef2835e7742c866b4c7c0406754376c2c8285a36c630346aa5c5f833@holesky.aestus.live
Ultrasound: https://0xb1559beef7b5ba3127485bbbb090362d9f497ba64e177ee2c8e7db74746306efad687f2cf8574e38d70067d40ef136dc@relay-stag.ultrasound.money

--------------
MEV-Boost holesky DNP repo: https://github.com/dappnode/DAppNodePackage-mev-boost-holesky

PR #1898, opened 2024-03-22
Add MEV-Boost package to Stakers Holesky

GiselleNessi:
Please see issue #1865

This pull request adds support for the Holesky testnet to MEV-Boost, including integration of working relays for Flashbots, Aestus, and Ultrasound networks. Additionally, it ensures a consistent user experience across different networks, mirroring the UX of MEV-Boost on other platforms. With Prater nearing deprecation and its relays no longer maintained, this update is crucial for seamless transition and continued functionality.

### DIFF
--- a/packages/admin-ui/src/pages/stakers/components/StakerNetwork.tsx
+++ b/packages/admin-ui/src/pages/stakers/components/StakerNetwork.tsx
@@ -233,7 +233,7 @@ export default function StakerNetwork<T extends Network>({
                   isSelected={newEnableWeb3signer}
                 />
               </Col>
-              {(network === "prater" || network === "mainnet") && (
+              {(network === "prater" || network === "mainnet" || network === "holesky") && (
                 <Col>
                   <SubTitle>Mev Boost</SubTitle>
                   <MevBoost
--- a/packages/admin-ui/src/pages/stakers/components/StakerNetwork.tsx
+++ b/packages/admin-ui/src/pages/stakers/components/StakerNetwork.tsx
@@ -233,7 +233,7 @@ export default function StakerNetwork<T extends Network>({
                   isSelected={newEnableWeb3signer}
                 />
               </Col>
-              {(network === "prater" || network === "mainnet" || network === "holesky") && (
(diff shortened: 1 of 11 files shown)


packages/stakers/src/set/setMevBoost.ts
```diff
@@ -22,7 +23,7 @@ export async function setMevBoost<T extends Network>({
   currentMevBoostPkg,
 }: {
   dappnodeInstaller: DappnodeInstaller;
-  mevBoost: T extends "mainnet" ? MevBoostMainnet : MevBoostPrater;
+  mevBoost: T extends "mainnet" ? MevBoostMainnet : MevBoostPrater | MevBoostHolesky;
```
dappnodedev, line 26: This looks incorrect, it would be:
`T extends "mainnet" ? MevBoostMainnet : T extends"prater" ? MevBoostPrater : MevBoostHolesky;`
or something similar

dappnodedev (changes requested):

packages/admin-ui/src/pages/stakers/components/StakerNetwork.tsx
```diff
@@ -233,7 +233,7 @@ export default function StakerNetwork<T extends Network>({
                   isSelected={newEnableWeb3signer}
                 />
               </Col>
-              {(network === "prater" || network === "mainnet") && (
+              {(network === "prater" || network === "mainnet" || network === "holesky") && (
```
dappnodedev, line 236: Probably there is a better way to do this. Something like:
`["prater", "mainnet", "holesky"].includes(network)`

(research a bit on the best way to do it)

GiselleNessi: The use of `Array.includes()` makes the condition more expressive and self-explanatory. It clearly communicates the intent of checking whether the network value is among a predefined list of valid networks. So it is a better approach for readability, maintainability, and flexibility of the code.

Marketen (approved):

PR #1903, opened 2024-04-03
Updated telegram id regex to allow for groups ids

GiselleNessi:
Consolidated email and Telegram channel ID validations into a single useEffect hook for improved code organization and readability. Additionally, enhanced the regex pattern to allow for groups in the Telegram channel ID format, providing added flexibility and versatility.

Telegram IDs can be in the format `'-1001234567890'` or `'-1001761649725_393'`


pablomendezroyo:
Already implemented

PR #1906, opened 2024-04-03
Update descriptions of chains in stakers UI

alexpeterson91:
## Context
The descriptions for the various blockchains we support in the stakers UI are... terrible.  
I rewrote them to make them actually relevant and explain things about the network on its goals/abilities, not just like `ethereum is based on blockchain tech, it has a programming language called solidity` 

## Approach

For the sources of the new descriptions I used references from each of the projects' home page/about descriptions.
I found it hard to even fit single sentences for each since the description just seemed too, incomplete or like it needed another phrase, so after checking with Vercel CI that having 2-3 sentences for each description doesn't make the display appear weird or off to me (though i do request whoever reviews this to give me their feedback not only on my changes to the descriptions but also whether or not this would look better if the page were modified to center justify the descriptions, i dont think its necessary, but its not hard to do, so I'd also like feedback on how it looks and not just how it reads.) 
So once i determined that at least I personally dont think any description is too long i took the info i had written myself from the project pages and then used Chat GPT4 to generate the best descriptions given my prompts and its task to describe the networks in 2-3 clear, concise sentences.  I tweaked what it came back with with my own writing style and added a phrase or 2 to each before determining that i think it's complete now.

## Test instructions

The only test is to open and view the vercel deployment from this PR, or re deploy it if its not active when you review.

I'm looking for 2 things, 

1. feedback on the descriptions themselves, if theyre good, too long, too wordy, missing info, wrong info[I hope not],  too much info etc.
2. and secondly feedback on how the descriptions look in the stakersUI and if you think it looks fine, or should be improved by editing the page to center justify the div block containing these descriptions on each tab.

### DIFF
--- a/packages/admin-ui/src/pages/stakers/components/StakersRoot.tsx
+++ b/packages/admin-ui/src/pages/stakers/components/StakersRoot.tsx
@@ -17,7 +17,7 @@ const StakersRoot: React.FC = () => {
           StakerNetwork({
             network: "mainnet",
             description:
-              "Ethereum is an open source, distributed software platform that is based on blockchain technology. It has its own native cryptocurrency called Ether and a programming language called Solidity."
+              "Ethereum is a decentralized, permissionless blockchain platform that employs a proof-of-stake (PoS) consensus mechanism, where validators stake Ether (ETH, native token of Ethereum) to secure the network and validate transactions in exchange for ETH rewards. This approach enhances network security and energy efficiency while also supporting the development and execution of smart contracts enabling robust decentralized applications (dApps) creating endless opportunities in its open-source ecosystem."
           })
       },
       {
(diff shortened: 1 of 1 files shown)


packages/admin-ui/src/pages/stakers/components/StakersRoot.tsx
```diff
@@ -47,7 +47,7 @@ const StakersRoot: React.FC = () => {
           StakerNetwork({
             network: "prater",
             description:
-              "The resulting testnet from the Prater and Göerli merge is the long-standing Ethereum testnet. Node operators can use it to test their node setups and app developers to test their stack"
+              "The resulting testnet from the Prater and Göerli merge is the long-standing Ethereum testnet. Node operators can use it to test their node setups and app developers to test their stack before deploying to the Mainnet."
```
alexpeterson91, line 50:
```suggestion
              "Göerli, the proper name for the resulting testnet created from the Prater and Göerli merge, is the long-standing Ethereum testnet. Node operators can use it to test their node setups and app developers to test their stack before deploying to the Mainnet.  It has since been replaced by the Holesky Testnet that was launched in a merged state and did not need to merge EL and CL layers as was done in Göerli"
```

pablomendezroyo (approved):

PR #1908, opened 2024-04-04
Ethical Metrics fixes before publish

mateumiralles:
- Disabling Switches while loading.
- Allowing to turn `OFF` the switch even if its fields are empty.
- Checking API status and don't allow the register until its status is up.

### DIFF
--- a/packages/dappmanager/src/calls/ethicalMetrics.ts
+++ b/packages/dappmanager/src/calls/ethicalMetrics.ts
@@ -8,8 +8,7 @@ import { logs } from "@dappnode/logger";
 import {
   ethicalMetricsDnpName,
   register,
-  unregister,
-  isApiUp
+  unregister
 } from "@dappnode/ethicalmetrics";
 import { dockerContainerStart, dockerContainerStop } from "@dappnode/dockerapi";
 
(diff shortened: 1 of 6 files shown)


packages/dappmanager/src/calls/ethicalMetrics.ts
```diff
@@ -78,6 +79,14 @@ export async function enableEthicalMetrics({
       if (!container.running)
         await dockerContainerStart(container.containerName);
 
+    //Checking if Ethical Metrics API is up before registering
+    let isEthicalUp = await isApiUp();
+    while (!isEthicalUp) {
```
pablomendezroyo, line 84: this might cause a potential infinite loop issue because setInterval is non-blocking and schedules an asynchronous operation, the loop does not wait for isApiUp() to complete before continuing. This means the while loop could potentially create an enormous number of intervals in a very short time

pablomendezroyo: avoid using while,

packages/ethicalMetrics/src/isApiUp.ts
```diff
@@ -0,0 +1,13 @@
+import * as url from "url";
+import fetch from "node-fetch";
+export async function isApiUp(): Promise<boolean> {
```
pablomendezroyo, line 3: I would instead use this function internally in the ethical metrics module on every API request.

You could add a function parameter for the timeout to wait before determining is up or not

pablomendezroyo (changes requested):

pablomendezroyo:
Closed by https://github.com/dappnode/DNP_DAPPMANAGER/pull/1914

PR #1917, opened 2024-04-09
Start container before docker cp

dappnodedev:
Before copying files to a docker container using `docker cp` the container needs to be running. This causes trouble in packages like Obol, where some services are stopped by default


dappnodedev:
Should we stop the container after copying the files or keep it running? For Obol specifically, probably it makes more sense to keep it running, but let's make it as generic as possible.

pablomendezroyo (approved):

3alpha (approved):

PR #1918, opened 2024-04-10
Add telegram user ID and dappnode AWS cloud

pablomendezroyo:
Add telegram user ID and use to only allow commands from this user for authenticated commands such as `/get_wireguard_credentials`

Also implement a AWS cloud fetcher to determine if its a dappnode cloud or not


packages/db/src/system.ts
```diff
@@ -8,6 +8,7 @@ const PASSWORD_IS_SECURE = "password-is-secure";
 const VERSION_DATA = "version-data";
 const TELEGRAM_STATUS = "telegram-status";
 const TELEGRAM_TOKEN = "telegram-token";
+const TELEGRA_USER_ID = "telegram-user-id";
```
3alpha, line 11:
```suggestion
const TELEGRAM_USER_ID = "telegram-user-id";
```

packages/db/src/system.ts
```diff
@@ -49,6 +50,10 @@ export const telegramChannelIds = dbMain.staticKey<string[]>(
   []
 );
 
+// Telegram user Id
+
+export const telegramUserId = dbMain.staticKey<string>(TELEGRA_USER_ID, "");
```
3alpha, line 55:
```suggestion
export const telegramUserId = dbMain.staticKey<string>(TELEGRAM_USER_ID, "");
```

3alpha (changes requested):

packages/daemons/src/telegramBot/dappnodeTelegramBot.ts
```diff
@@ -63,6 +73,27 @@ export class DappnodeTelegramBot {
...
+          if (!userId) return;
+          if (userId.toString() !== db.telegramUserId.get()) {
+            logs.error(`Unauthorized user: ${userId}`);
+            await this.sendMessage(
+              msg.chat.id.toString(),
+              "Unauthorized user. Please contact the admin"
+            );
+          } else {
+            // default device https://github.com/dappnode/DNP_WIREGUARD/blob/4a074010c98b5d3003d1c3306edcb75392b247f4/docker-compose.yml#L12
+            const deviceName = msg.text.split(" ")[1] || "dappnode_admin";
```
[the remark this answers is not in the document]
pablomendezroyo: like what

packages/daemons/src/telegramBot/dappnodeTelegramBot.ts
```diff
@@ -90,10 +121,32 @@ export class DappnodeTelegramBot {
...
+      parse_mode: "Markdown",
     });
   }
 
+  // Utils
+
+  /**
+   *  Fetch wireguard credentials with the default device
+   */
+  private async getWireguardCredentialsCmd(
```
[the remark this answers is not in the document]
pablomendezroyo: true, but the src code is in the  dappmanager module. Ideal solution should be create a separated module for wireguard and use it either in dappmanager and here

packages/admin-ui/src/pages/system/components/Notifications/Telegram.tsx
```diff
@@ -135,8 +142,110 @@ export function TelegramNotifications() {
...
+          }
+        />
+        {tokenError && (
+          <span style={{ fontSize: "12px", color: "red" }}>
+            Telegram token format is incorrect
+          </span>
+        )}
+        <br />
+        <Form.Label>Telegram user ID</Form.Label>
+        <Input
```
mateumiralles, line 174: Using `<InputSecret/>` instead of `<Input/>`?

PR #1927, opened 2024-04-16
Create staker network

pablomendezroyo:
Create staker network for staker pkgs engine communications: `staker_network`
- It will have an initial migration where it will connect selected consensus and execution clients from every network to the `staker_network` as well as writing the corresponding compose configuration in the `docker-compose.yml` file. It will also add a fullnode alias to the execution client, to be used by the consensus client in the engine communication
- It will take the same approach on every user selection in the stakers
- Furthermore, it will attempt to disconnect any other client from the stakers and remove the corresponding configuration from the compose file for reliability.

**Benefits**:
- More flexibility when adding new execution/consensus client. There will not be needed a release of every client allowing this new client, but just a core release
- Stop from using global envs: this will have an impact of stopping from restarting packages that use global envs to realize whenever its used a specific execution/consensus client. Instead, it will be used a single unique domain.
- Clients pkgs entrypoints simplification

**Next steps**
- Test
- Use single JWT per network
- Use fullnode domain in consensus clients


packages/migrations/src/createStakerNetworkAndConnectStakerPkgs.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Creates the docker staker network
+ */
+async function createDockerStakerNetwork(): Promise<void> {
+  try {
+    const stakerNetwork = docker.getNetwork(params.DOCKER_STAKER_NETWORK_NAME);
+    await stakerNetwork.inspect();
+    logs.info(`docker network ${params.DOCKER_STAKER_NETWORK_NAME} exists`);
+  } catch (e) {
+    if (e.statusCode === 404) {
+      logs.info(`Creating docker network ${params.DOCKER_STAKER_NETWORK_NAME}`);
```
Marketen, line 24: I would maybe add "Docker network `params.DOCKER_STAKER_NETWORK_NAME` not found, creating it" as a debug log to add more context.

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
```diff
@@ -0,0 +1,194 @@
...
+      if (!dnpName) continue;
+
+      const pkg = await listPackageNoThrow({
+        dnpName: dnpName,
+      });
+      // skip if pkg not installed
+      if (!pkg) continue;
+
+      // Disconnect and remove staker network from other clients
+      await disconnectAndRemoveStakerNetworkFromCompose(clientsToCheck);
```
Marketen, line 50: why is there no error handling here or in `disconnectAndRemoveStakerNetworkFromCompose()` itself? calling a container to disconnect can return "server error" or "not found" error: https://docs.docker.com/engine/api/v1.37/#tag/Network/operation/NetworkDisconnect

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * Connects the staker pkg to the staker network with the fullnode alias
+ */
+async function connectPkgToStakerNetwork(
+  networkName: string,
+  pkgContainers: PackageContainer[],
+  alias?: string | null
+): Promise<void> {
+  for (const container of pkgContainers)
+    await dockerNetworkConnectNotThrow(networkName, container.containerName, {
```
Marketen, line 175: same as above, why use notThrow and no error handling? being successfully connected tot his new staker network seems important enough to check it

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+}
+
+/**
+ * Connects the staker pkg to the staker network with the fullnode alias
+ */
+async function connectPkgToStakerNetwork(
+  networkName: string,
+  pkgContainers: PackageContainer[],
+  alias?: string | null
```
Marketen, line 172: should we ensure alias is not null if package to connect is an execution client?

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
pablomendezroyo: we do not want to move forward if the docker compose config has not bee removed successfully, otherwise it may lead to more than exec/cons client from the same network in the staker_network

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
pablomendezroyo: yeah better to only call the throwable function but with the disconnected containers

packages/migrations/src/createStakerNetworkAndConnectStakerPkgs.ts
```diff
@@ -0,0 +1,42 @@
+import { docker } from "@dappnode/dockerapi";
+import { params } from "@dappnode/params";
+import { logs } from "@dappnode/logger";
+import { ensureStakerPkgsNetworkConfig } from "@dappnode/stakers";
+import { networks } from "@dappnode/types";
+
+/**
+ * Creates the staker network and connects the staker packages to it
+ */
+export async function createStakerNetworkAndConnectStakerPkgs(): Promise<void> {
```
dappnodedev, line 10: In general, functions should have a single purpose. Including "and" in a function name is indicating something is wrong. 

Maybe they can be called independently or creating the staker network could be a pre requisite

packages/migrations/src/createStakerNetworkAndConnectStakerPkgs.ts
```diff
@@ -0,0 +1,42 @@
...
+  for (const network of networks) await ensureStakerPkgsNetworkConfig(network);
+}
+
+/**
+ * Creates the docker staker network
+ */
+async function createDockerStakerNetwork(): Promise<void> {
+  try {
+    const stakerNetwork = docker.getNetwork(params.DOCKER_STAKER_NETWORK_NAME);
+    await stakerNetwork.inspect();
```
dappnodedev, line 21: Why calling here inspect?

dappnodedev: Is it necessary for the error to be thrown?

dappnodedev: Can't you just check if the network exists and then create it if not? Instead of triggering an error for something that is expected

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
```diff
@@ -0,0 +1,202 @@
...
+ * - **selected** consensus client
+ */
+export async function ensureStakerPkgsNetworkConfig(
+  network: Network
+): Promise<void> {
+  const stakerConfig = getStakerConfigByNetwork(network);
+  const { executionClients, consensusClients } =
+    getStakerDnpNamesByNetwork(network);
+
+  for (const { dnpName, clientsToCheck, alias } of [
```
dappnodedev, line 29: I think this for loop is quite complex, can you try to make it clearer to read?

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
```diff
@@ -0,0 +1,202 @@
...
+      [serviceName].removeNetwork(params.DOCKER_STAKER_NETWORK_NAME);
+    composeService.write();
+  }
+}
+
+async function disconnectConnectedPkgFromStakerNetwork(
+  networkName: string,
+  pkgContainers: PackageContainer[]
+): Promise<void> {
+  const connectedContainers = pkgContainers
```
dappnodedev, line 112: Not related to this exact line, but we need to be sure that a container connected to the stakers network that becomes stopped will not re-connect once it is started

dappnodedev (changes requested):

packages/stakers/src/set/ensureStakerPkgsNetworkConfig.ts
pablomendezroyo: thats ensure with the compose file

packages/stakers/src/stakerComponent.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    // start all containers
+    //if (!pkg.containers.some((c) => c.running))
+    await dockerComposeUpPackage({ dnpName: pkg.dnpName }, true);
+  }
+
+  /**
+   * Adds the staker network and its fullnode alias to the docker-compose file
+   */
+  private addStakerNetworkToCompose(
```
dappnodedev, line 176: What do you think about something like:

```  /**
   * Adds the staker network and its fullnode alias to the docker-compose file
   */
  private addStakerNetworkToCompose(
    dnpName: string,
    netConfigsToAdd: { [serviceName: string]: ComposeServiceNetworksObj }
  ): void {
    const composeEditor = new ComposeFileEditor(dnpName, false);
    const services = composeEditor.compose.services;
    const rootNetworks = composeEditor.compose.networks || {};

    for (const [serviceName, networkConfig] of Object.entries(netConfigsToAdd)) {

      // Find the service that includes serviceName in its name
      const service = services[serviceName] ||
        Object.entries(services).find(([name]) => name.includes(serviceName))?.[1];

      if (!service) {
        logs.warn(`Service ${serviceName} not found in ${dnpName}, skipping`);
        continue; // TODO: Throw error?
      }

      const serviceNetworks = services[serviceName].networks || {};

      // Merge the network config with the existing service networks
      service.networks = {
        ...serviceNetworks,
        ...networkConfig
      };

      // Ensure all networks are added to the root level
      for (const [networkName] of Object.entries(networkConfig)) {
        if (!rootNetworks[networkName]) {
          rootNetworks[networkName] = {
            external: true
          };
        }
      }
    }

    composeEditor.compose.networks = rootNetworks;

    composeEditor.write();
  }

```

dappnodedev: Note that here instead of using a regex I just try to find the exact service name and, if not found, I look for the service that includes that name

dappnodedev: 70 lines vs 40 lines

dappnodedev (changes requested):

packages/stakers/src/stakerComponent.ts
```diff
@@ -0,0 +1,356 @@
...
+    const connectedContainers = pkgContainers
+      .filter((container) =>
+        container.networks.some((network) => network.name === networkName)
+      )
+      .map((container) => container.containerName);
+    for (const container of connectedContainers)
+      await dockerNetworkDisconnect(networkName, container);
+  }
+
+  private removeStakerNetworkFromCompose(
```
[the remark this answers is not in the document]
dappnodedev: But we can log a warning or throw an error for array case if needed (can't add aliases)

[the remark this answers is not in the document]
pablomendezroyo: committed suggestion in latest commit

packages/stakers/src/stakerComponent.ts
pablomendezroyo: committed suggestion see https://github.com/dappnode/DNP_DAPPMANAGER/pull/1927/commits/a2d6ebace974acf48763a0672f53737ff12d4a33

PR #1932, opened 2024-04-18
Improve upstream repo schema

dappnodedev:
Improve upstream repo schema to support the old way of handling upstream repos together with the newly defined in https://github.com/dappnode/DAppNodeSDK/issues/408

There are 3 possible cases:
1. No upstream settings defined
2. Upstream settings defined in the old way (upstreamRepo, upstreamArg and upstreamVersion fields)
3. Upstream settings defined in the new way (upstream object, with the fields repo, version and arg inside it)

It is not possible to define both at the same time


dappnodedev:
Needs some testing

pablomendezroyo (approved):

pablomendezroyo (approved):

PR #1933, opened 2024-04-19
Remove swap card

dappnodedev:
Removed swap card, as it is causing confusion among the users


pablomendezroyo (changes requested):
Remove also the backend call

PR #1936, opened 2024-04-19
Only support 1 upstream repo with separate fields

dappnodedev:
After researching, I have not found any package that defines multiple upstream repos like an array of string, so we have decided that the only way of having multiple upstream projects will be using the new format defined in https://github.com/dappnode/DAppNodeSDK/issues/408#issuecomment-2063702185


pablomendezroyo (approved):
Nice code quality PR :)

PR #1946, opened 2024-05-13
Prevent packages routes from auto-rendering

mateumiralles:
Preventing the package tab routes from auto-reloading after cursor being focused in another window. 
Passing the `react.render` as a `JSX element`, was creating a new instance of the element on each render, triggering unnecessary re-renders.


dappnodedev (approved):

dappnodedev:
Worked for me, nice catch!

PR #1964, opened 2024-06-17
Implement requirement not installed packages

pablomendezroyo:
Implement requirement not installed packages. This requirement will enforce to uninstall a package manually before moving forward with an update or a installation from scratch


dappnodedev (approved):

dappnodedev (changes requested):
Tests are failing

dappnodedev:
I approve the approach, but maybe it would be better UX to automatically delete the packages specified to be uninstalled when installing the new one. Obviously, we would need to show the user an alert related to this before proceeding with the installation.

Also, I am curious to know what will happen if auto updates are triggered for package A, but package B needs to be uninstalled first. Will it be installed anyway? Will the auto update be stopped?

PR #1967, opened 2024-06-18
Implement auth for https in backend

pablomendezroyo:
Implement auth for https in backend:

- Allows to add optional query argument to add mapping API call **auth**
- The auth comes as a string in format `${username}:{SSHA}${combined}`
- No password stored in dappmanager side


3alpha (approved):

pablomendezroyo:
> It needs to be clear for the user that setting user and pass when creating a new mapping is completely optional

button gets enabled just with from domain and port

3alpha (approved):

PR #1974, opened 2024-07-02
Fix persist docker networks

pablomendezroyo:
Fix persist docker networks and user settings related to docker networks

### DIFF
--- a/packages/stakers/src/consensus.ts
+++ b/packages/stakers/src/consensus.ts
@@ -108,11 +108,13 @@ export class Consensus extends StakerComponent {
       newStakerDnpName: newConsensusDnpName,
       dockerNetworkName: params.DOCKER_STAKER_NETWORKS[network],
       compatibleClients: Consensus.CompatibleConsensus[network],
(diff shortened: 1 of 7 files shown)


packages/stakers/src/consensus.ts
```diff
@@ -90,84 +91,142 @@ export class Consensus extends StakerComponent {
...
     await super.setNew({
       newStakerDnpName: newConsensusDnpName,
       dockerNetworkName: params.DOCKER_STAKER_NETWORKS[network],
       compatibleClients: Consensus.CompatibleConsensus[network],
-      dockerNetworkConfigsToAdd: this.getNetworkConfigsToAdd(network),
-      userSettings,
+      userSettings: newConsensusDnpName
+        ? this.getUserSettings(
+            newConsensusDnpName,
+            Boolean(await listPackageNoThrow({ dnpName: newConsensusDnpName })),
```
dappnodedev, line 121: I would create an `isInstalled` const to make this cleaner

PR #1975, opened 2024-07-02
Fix `getPkgAsset` to return an array if multiple elements exist

Marketen:
closes https://github.com/dappnode/DAppNodePackage-DMS/issues/47 as part of the project https://github.com/orgs/dappnode/projects/72/views/1

**getPkgAsset()** now takes into account the "multiple" field of `fileConfig` parameter. If `multiple == true` and at least one file (here defined as IpfsEntry) passes the given name regex for the `fileConfig`,  **getPkgAsset()** will return `<T>` as a **json array**, instead of a **json object**.

For example, [dappnode exporter package](https://github.com/dappnode/DAppNodePackage-exporter) has two files that end with `grafana-dashboard.json`. When the IPFS content of the package is downloaded, dappmanager will now notice that both files must be appended as an array, since we permit multiple `grafana-dashboard.json` to exist in a dnp, and both files pass the `grafana-dashboard.json` regex

Returning the content in grafana-dashboards.json as an array solves the DMS bug, since DMS's manager expects this content to be an array

### DIFF
--- a/packages/toolkit/src/repository/repository.ts
+++ b/packages/toolkit/src/repository/repository.ts
@@ -254,37 +254,32 @@ export class DappnodeRepository extends ApmRepository {
  * @throws - If the file is required and not found.
  * @returns - The release package for the request or undefined if not found.
  */
-public async getPkgAsset<T>(
-  fileConfig: FileConfig,
-  ipfsEntries: IPFSEntry[]
-): Promise<T | undefined> {
-  const { regex, required, multiple } = fileConfig;
-
-  // We filter the entries (files) so that we only consider the ones that match the regex.
-  // for example, all grafana dashboards must pass /.*grafana-dashboard.json$/ regex
-  const matchingEntries = ipfsEntries.filter(file => regex.test(file.name));
-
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/repository/repository.ts
```diff
@@ -556,33 +563,42 @@ export class DappnodeRepository extends ApmRepository {
...
+        default: {
+          throw new Error(`Unsupported format: ${format}`);
+        }
+      }
+    };
+  
+    try {
+      if (Array.isArray(data)) {
+        // Map over array data if it's an array, using parseSingle for each item
+        return data.map(parseSingle) as unknown as T;
```
pablomendezroyo, line 590: Using `as unknown` is considered a really bad practice in ts. Is there any other workaround in this approach?

pablomendezroyo (changes requested):

packages/toolkit/src/repository/repository.ts
Marketen: fixed, removed "unknown" and "any" types.

pablomendezroyo (approved):

PR #2007, opened 2024-08-05
Use compose plugin for file validation

dappnodedev:
Use new `compose` plugin for file validation. instead of legacy binary `docker-compose`


dappnodedev:
Tests are failing due to the bug described in this PR

pablomendezroyo (approved):

PR #2011, opened 2024-08-12
Allow semver ranges

dappnodedev:
This PR refactors the `DappGetFetcher` class to improve the handling of dependencies and their version ranges. The main improvements include:

1. **Separation of Concerns:**
   - The `processDependencies` method was introduced to handle two main responsibilities:
     - Filtering out dependencies that are already satisfied by installed packages.
     - Parsing and converting semver ranges to appropriate APM-compatible versions.

2. **Improved Optional Dependencies Handling:**
   - A new `mergeOptionalDependencies` method was introduced to inject optional dependencies into the main dependencies object only if the corresponding packages are installed.

3. **Enhanced Semver Range Handling:**
   - The `parseSemverRangeToApmVersion` method was created to convert semver ranges like `^x.x.x`, `~x.x.x`, `>x.x.x`, and `>=x.x.x` to appropriate APM-compatible versions or the latest version (`*`) if applicable.
   - Added checks to ensure that only simple version ranges are supported, throwing errors for complex or combined ranges.

4. **Code Readability and Maintainability:**
   - The code was refactored to be more modular and easier to understand, with clear method responsibilities and added inline documentation.
   - Methods were renamed and reorganized for clarity, following best practices.

### DIFF
--- a/packages/installer/src/dappGet/fetch/DappGetFetcher.ts
+++ b/packages/installer/src/dappGet/fetch/DappGetFetcher.ts
@@ -130,16 +130,39 @@ export class DappGetFetcher {
   }
 
   /**
- * Resolves and updates the given dependencies to their exact version by determining the maximum satisfying version.
- * 
- * This method takes a set of dependencies where the version is specified as a semver range and resolves it to the exact
- * version that should be installed. It does so by fetching all published versions of each dependency from the APM and 
- * determining the highest version that satisfies the given semver range.
- *
- * @param dependencies - An object representing the dependencies where the key is the dependency name and the value is the semver range or version.
- * @param dappnodeInstaller - An instance of `DappnodeInstaller` used to fetch published versions from the APM.
- * @throws If a semver range is invalid or if no satisfying version can be found for a dependency.
- */
+   * Resolves and updates the given dependencies to their exact version by determining the maximum satisfying version.
+   * 
+   * This method takes a set of dependencies where the version is specified as a semver range and resolves it to the exact
+   * version that should be installed. It does so by fetching all published versions of each dependency from the APM and 
+   * determining the highest version that satisfies the given semver range.
+   *
+   * @param dependencies - An object representing the dependencies where the key is the dependency name and the value is the semver range or version.
+   * @param dappnodeInstaller - An instance of `DappnodeInstaller` used to fetch published versions from the APM.
+   * @throws If a semver range is invalid or if no satisfying version can be found for a dependency.
+   * 
+   * @example
+   * // Given the following dependencies object with semver ranges:
+   * const dependencies = {
+   *   "example-dnp": "^1.0.0",
+   *   "another-dnp": "2.x",
+   *   "ipfs-dnp": "/ipfs/QmXf2...abc"
+   * };
+   * 
+   * // And assuming the following versions are available in the APM:
+   * // example-dnp: ["1.0.0", "1.1.0", "1.2.0"]
+   * // another-dnp: ["2.0.0", "2.1.0", "2.5.0"]
+   * // ipfs-dnp: ["/ipfs/QmXf2...abc"]
+   * 
+   * // After calling defineExactVersions:
+   * await defineExactVersions(dependencies, dappnodeInstaller);
+   * 
+   * // The dependencies object will be updated to:
+   * // {
+   * //   "example-dnp": "1.2.0",  // The highest version satisfying "^1.0.0"
+   * //   "another-dnp": "2.5.0",  // The highest version satisfying "2.x"
+   * //   "ipfs-dnp": "/ipfs/QmXf2...abc"  // Exact match
+   * // }
+   */
   private async defineExactVersions(
     dependencies: Dependencies,
     dappnodeInstaller: DappnodeInstaller
--- a/packages/installer/src/dappGet/fetch/DappGetFetcher.ts
+++ b/packages/installer/src/dappGet/fetch/DappGetFetcher.ts
@@ -2,6 +2,7 @@ import { Dependencies, InstalledPackageData } from "@dappnode/types";
 import { validRange, satisfies, valid, maxSatisfying } from "semver";
 import { DappnodeInstaller } from "../../dappnodeInstaller.js";
 import { listPackages } from "@dappnode/dockerapi";
+import { isIpfsHash } from "../../utils.js";
 
 export class DappGetFetcher {
   /**
@@ -10,7 +11,7 @@ export class DappGetFetcher {
    * @returns dependencies:
    *   { dnp-name-1: "semverRange", dnp-name-2: "/ipfs/Qmf53..."}
    */
-  async dependencies(
+  async dependenciesToInstall(
     dappnodeInstaller: DappnodeInstaller,
     name: string,
     version: string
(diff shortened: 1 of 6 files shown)


packages/installer/src/dappGet/fetch/DappGetFetcher.ts
```diff
@@ -78,7 +68,93 @@ export class DappGetFetcher {
...
+
+  /**
+   * Merges optional dependencies into the main dependencies object if the corresponding 
+   * packages are installed.
+   *
+   * @param dependencies The main dependencies object to be merged into.
+   * @param optionalDependencies The optional dependencies to be checked and merged.
+   * @param installedPackages The list of currently installed packages.
+   */
+  private mergeOptionalDependencies(
```
pablomendezroyo, line 83: I like it, much cleaner

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
```diff
@@ -78,7 +68,93 @@ export class DappGetFetcher {
...
+
+  /**
+   * Parses a semver range and converts it to a version suitable for APM.
+   * Handles specific semver operators like ^, ~, >, >= and converts them to
+   * appropriate APM-compatible versions.
+   *
+   * @param semverRange The semver range to be parsed.
+   * @returns The parsed version as a string.
+   */
+  private parseSemverRangeToApmVersion(semverRange: string): string {
```
pablomendezroyo, line 143: For context: 

- The current package manifest version pattern is https://github.com/dappnode/DNP_DAPPMANAGER/blob/bafd376cf6b233da40e9d163477c02fa38adeec5/packages/schemas/src/schemas/manifest.schema.json#L42
```json
"pattern": "^((([0-9]+).([0-9]+).([0-9]+)))$",
```

- The repository SC only allows strict semver, see https://github.com/dappnode/DNP_DAPPMANAGER/blob/bafd376cf6b233da40e9d163477c02fa38adeec5/packages/toolkit/truffle/contracts/contracts_v0.4/repository.sol#L503
```solidity
    function isValidBump(uint16[3] _oldVersion, uint16[3] _newVersion) public pure returns (bool) {
        bool hasBumped;
        uint i = 0;
        while (i < 3) {
            if (hasBumped) {
                if (_newVersion[i] != 0) {
                    return false;
                }
            } else if (_newVersion[i] != _oldVersion[i]) {
                if (_oldVersion[i] > _newVersion[i] || _newVersion[i] - _oldVersion[i] != 1) {
                    return false;
                }
                hasBumped = true;
            }
            i++;
        }
        return hasBumped;
    }
```

Which means that we will kind of follow JS ecosystem where packages will have always a semver version and the package manager can parse the dependency target defined in the manifest (package.json)

pablomendezroyo: we should document this in the dev documentation letting devs know that they are allowed to do so

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
```diff
@@ -78,7 +68,93 @@ export class DappGetFetcher {
...
+    if (/\d+\.\d+\.\d+$/.test(semverRange))
+      return semverRange;
+
+    // Use "*" (latest) if version is >... or >=...
+    if (/^>=?\d+\.\d+\.\d+$/.test(semverRange)) {
+      return '*';
+    }
+
+    // Use x.x.x if version is ^x.x.x or ~x.x.x
+    if (/^[\^~]\d+\.\d+\.\d+$/.test(semverRange)) {
```
pablomendezroyo, line 154: is this correct? shouldnt determine the following?
```
^1.2.3 - Matches the latest patch version for the 1.2.x range.
~1.2.3 - Matches the latest minor version for the 1.x.x range.
```

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
pablomendezroyo, line 143: I would rename the function to something like `parseSemverRangeToSemverStrict` since semver is something not specific to APM

pablomendezroyo (changes requested):

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
dappnodedev: This has been fixed in https://github.com/dappnode/DNP_DAPPMANAGER/pull/2011/commits/2b8c470724926416d5deb8ece8adba932811e240

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
dappnodedev: This function has been removed, as it is not needed anymore

dappnodedev: Actually, I am able to generate and to install packages with semver ranges that are not considered in that pattern

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
```diff
@@ -78,7 +71,85 @@ export class DappGetFetcher {
...
+  }
+
+  /**
+   * Processes the dependencies by first filtering out those that are already satisfied by installed packages
+   * and then converting any remaining semver ranges to appropriate APM versions.
+   *
+   * @param dependencies The main dependencies object to be processed.
+   * @param installedPackages The list of currently installed packages.
+   */
+  private filterSatisfiedDependencies(
```
pablomendezroyo, line 109: It seems to me weird deleting the dependency if its installed already. It must be checked still if it satisfies the version range defined

pablomendezroyo: I would take a different approach than deleting the dependency, it is still a dependency so whenever its going to be installed that code should bypass if already installed and version satisfies

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
```diff
@@ -78,7 +71,85 @@ export class DappGetFetcher {
...
+  }
+
+  private async defineExactVersions(
+    dependencies: Dependencies,
+    dappnodeInstaller: DappnodeInstaller
+  ): Promise<void> {
+
+    for (const [depName, depVersion] of Object.entries(dependencies)) {
+
+      if (validRange(depVersion)) {
```
pablomendezroyo, line 138: shouldnt be thwron an error if invalid range?

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
```diff
@@ -78,7 +71,85 @@ export class DappGetFetcher {
...
+      if (installedPackage && satisfies(installedPackage.version, depVersion)) {
+        console.log(
+          `Dependency ${depName} is already installed with version ${installedPackage.version}`
+        );
+        delete dependencies[depName];
+      }
+    }
+  }
+
+  private async defineExactVersions(
```
pablomendezroyo, line 131: could you add a description to the function also providing an example of how looks the object before and after?

pablomendezroyo (changes requested):

packages/installer/src/dappGet/fetch/DappGetFetcher.ts
dappnodedev: Yes! Done

dappnodedev:
This should not be necessary as it is already supported in another way

PR #2014, opened 2024-08-16
Prepare dappmanager for multi-service nimbus client

dappnodedev:
We are going to migrate from monoservice nimbus client to a new version that includes:

1. `beacon-chain`
2. `validator`
3. `beacon-validator` --> This is a proxy that forwards request following this pattern:

```
req to beacon-validator:4500 ----------> beacon-chain:3500
req to beacon-validator:3500 ----------> validator: 3500
````

This PR removes logic related to beacon-validator service

It is important to design a proper migration strategy, as publishing a new Dappmanager version including this changes will cause the previous versions of Nimbus client to fail if selected in the Stakers tab

### CODE REVIEW

File: packages/installer/src/ethClient/apiUrl.ts [hunk 1]
```diff
@@ -27,27 +27,28 @@ export function getEthExecClientApiUrl(dnpName: string, port = 8545): string {
...
-    port = dnp.chain.portNumber;
-    domain = buildNetworkAlias({
-      dnpName: dnpName,
-      serviceName: dnp.chain.serviceName,
-      isMainOrMonoservice: false
-    });
-  } else {
-    // Lighthouse, Teku and Prysm use 3500
-    // Nimbus uses 4500 because it is a monoservice and the validator API is using that port
-    if (dnpName.includes("nimbus")) {
```
dappnodedev, line 43: All production Nimbus clients have their driver defined like this:

```
  "chain": {
    "driver": "ethereum-beacon-chain",
    "serviceName": "beacon-validator",
    "portNumber": 4500
  },
  ```
  
  So it is not necessary to consider this condition. Also, once nimbus package is published from https://github.com/dappnode/DAppNodePackage-nimbus-generic , it will have 2 services that will behave the same as the rest of the clients

File: packages/utils/src/stakerUtils.ts [hunk 2]
```diff
@@ -1,70 +1,41 @@
...
-          : {
-              [validatorServiceName]: {
-                // Fee recipient is set as global env, keep this for backwards compatibility
-                ["FEE_RECIPIENT_ADDRESS"]: defaultFeeRecipient,
-                // Graffiti is a mandatory value
-                ["GRAFFITI"]: defaultDappnodeGraffiti
-              },
+    environment: {
+      // TODO: Remove once Nimbus is split into 2 services
+      [beaconValidatorServiceName]: {
```
dappnodedev, line 17: It is okay to include it in all cases, because the `applyUserSettings` function will only take into account the existing services in the compose of the package to install/start:
image

pablomendezroyo (approved):

pablomendezroyo (approved):

PR #2020, opened 2024-08-30
Improve port mapping warnings/errors

Marketen:
Dappmanager didnt let user map a new port if it was in the ephemeral range (32768-65535). These are ports used normally for a short period of time.

Now, only a warning is displayed if a port is in this range. An exception is made for 51820, which is the default Wireguard port, hardcoded by us.
See table where wireguard port 51820 is acknowledged: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports
DNP_WIREGUARD port config: https://github.com/dappnode/DNP_WIREGUARD/blob/4a074010c98b5d3003d1c3306edcb75392b247f4/docker-compose.yml#L17

Screenshot_20240830_143351

Also, if user tries to map a port > max port number (65535), the "update port mappings" button is disabled.

Screenshot_20240830_143307


packages/admin-ui/src/pages/packages/components/Network/PortsByService.tsx
```diff
@@ -120,59 +120,101 @@ export function PortsByService({
...
-  function getPortsOverTheMax(): PortMapping[] {
+  // Function to get any container ports in the ephemeral range, excluding port 51820
+  function getContainerPortsInEphemeralRange(): PortMapping[] {
     return ports.filter(
-      ({ host, container, deletable }) =>
-        (deletable &&
-          (container > maxPortNumber || (host && host > maxPortNumber))) ||
-        (host && host > maxEphemeralPortNumber)
+      ({ container, deletable }) =>
+        deletable && container > maxRegisteredPortNumber && container <= maxPortNumber && container !== 51820
```
dappnodedev, line 127: I would create a const `wireguardPort` where it is briefly explained, instead of hardcoding 51820 every time

Also, you could even create a function callled `isWireguardPort` which would be more self explanatory each time you check if port is 51820

packages/admin-ui/src/pages/packages/components/Network/PortsByService.tsx
```diff
@@ -120,59 +120,101 @@ export function PortsByService({
...
     );
   }
 
+  // Function to get any host ports in the ephemeral range, excluding port 51820
+  function getHostPortsInEphemeralRange(): PortMapping[] {
+    return ports.filter(
+      ({ host, deletable }) =>
+        deletable && host !== undefined && host > maxRegisteredPortNumber && host <= maxPortNumber && host !== 51820
```
dappnodedev, line 135: I would also create a function called `isInEphemeralRange` that would be called from `getContainerPortsInEphemeralRange` and `getHostPortsInEphemeralRange`

dappnodedev (approved):

3alpha (approved):

pablomendezroyo (approved):

Issue #1968, opened 2024-06-21
Ensure requirements in update time


pablomendezroyo:
Right now requirements are only enforeced in installation time, they must block an update if the requirement is not met.

Requirements:

- `minimumDappnodeVersion`
- `notInstalledPackages`
- `notInstalledPackages` (new see https://github.com/dappnode/DNP_DAPPMANAGER/pull/1964)

PR #2054, opened 2024-11-05
ensure requirements in `packageInstall` and not only on `fetchDnpRequest`

pablomendezroyo:
Ensure requirements in `packageInstall` and not only on `fetchDnpRequest`.

This will affect mainly auto updates as long as the install from the dappstore is already tracked with the compatible data


pablomendezroyo:
Closes https://github.com/dappnode/DNP_DAPPMANAGER/issues/1968

PR #2063, opened 2024-12-10
Add *.dappnode to cors

pablomendezroyo:
Add *.dappnode to cors to allow queries from dappnode UIs


packages/params/src/params.ts
```diff
@@ -86,7 +86,8 @@ export const params = {
     "http://localhost:3000",
     "http://localhost:3001",
     "http://my.dappnode",
```
3alpha, line 88:
```suggestion
```

As it is covered by wildcard.

PR #2094, opened 2025-03-07
Card style update

mateumiralles:
This PR enhances the DappManager UI with the following updates:
- Updated card styles by removing some borders, decreasing padding, and adjusting the border radius.
- Removed unnecessary cards that were wrapping entire tabs.


pablomendezroyo (changes requested):
Place text not in cards inside a card to improve redability

pablomendezroyo (approved):

PR #2226, opened 2025-08-20
add unhandled rejections net

Marketen:
prevents dappmanager from exiting if any daemon has an unhandled promise rejection


pablomendezroyo:
Already implemented

PR #2296, opened 2025-10-17
wait "removing" containers

Marketen:
This pull request introduces improvements to the `dockerComposeUpPackage` function in `dockerComposeUp.ts` to enhance reliability and prevent race conditions. The main changes are the addition of an in-process locking mechanism to prevent concurrent `docker-compose up` calls for the same compose file, and a check to block operations if any container is in a 'removing' state.

Concurrency control and reliability improvements:

* Added an in-process lock map (`composeUpLocks`) to prevent concurrent `dockerComposeUpPackage` executions for the same `composePath`, ensuring only one operation per compose file runs at a time. [[1]](diffhunk://#diff-54930a0b17a4e9dcfb1a4f5b418ed6277dd0064fc2358677e8397ddc89652e77R11-R13) [[2]](diffhunk://#diff-54930a0b17a4e9dcfb1a4f5b418ed6277dd0064fc2358677e8397ddc89652e77R42-R66) [[3]](diffhunk://#diff-54930a0b17a4e9dcfb1a4f5b418ed6277dd0064fc2358677e8397ddc89652e77R105-R112)
* Updated the function documentation to reflect the new concurrency prevention feature.

Container state validation:

* Added a check to prevent `dockerComposeUpPackage` from running if any service container is in the 'removing' state, improving error handling and preventing potential issues with partially removed containers.

### DIFF
--- a/packages/admin-ui/src/__mock-backend__/stakerConfig.ts
+++ b/packages/admin-ui/src/__mock-backend__/stakerConfig.ts
@@ -904,7 +904,7 @@ export const stakerConfig: Pick<Routes, "stakerConfigGet" | "stakerConfigSet"> =
           },
           feeRecipient: "0x71C7656EC7ab88b098defB751B7401B5f6d8976F"
         };
-      
+
       case "sepolia":
         return {
           executionClients: [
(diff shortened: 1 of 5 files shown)


packages/dockerApi/src/dockerComposeUp.ts
```diff
@@ -35,50 +39,76 @@ export async function dockerComposeUpPackage({
...
+    const timeout = containersStatus ? getDockerTimeoutMax(Object.values(containersStatus)) : undefined;
+    if (timeout && dockerComposeUpOptions && !dockerComposeUpOptions.timeout) dockerComposeUpOptions.timeout = timeout;
+
+    // Check for any service in 'removing' state before proceeding
+    const dnpData: InstalledPackageData | null = await listPackageNoThrow({ dnpName });
+    if (dnpData && dnpData.containers) {
+      const removingContainers = dnpData.containers.filter(c => c.state === "removing");
+      if (removingContainers.length > 0) {
+        const serviceNames = removingContainers.map(c => c.serviceName).join(", ");
+        logs.warn(`dockerComposeUpPackage: Cannot proceed, services in 'removing' state: ${serviceNames}`);
```
pablomendezroyo: I would remove warn log since u are already throwing and this error should be logged afterwards

PR #2336, opened 2026-01-02
Implement resilient IPFS fetcher

pablomendezroyo:
## Context

This pull request introduces a new mechanism for fetching content from IPFS using multiple gateway URLs, enhancing resilience and reliability in content retrieval.

## Approach

The solution involves modifying the IPFS repository structure to support an array of gateway URLs, implementing functions to set and get these URLs, and updating the metrics to reflect the number of configured gateways.

## Test instructions

1. Set multiple IPFS gateway URLs using the `ipfsGatewayUrlsSet` function.
2. Retrieve the configured URLs with the `ipfsGatewayUrlsGet` function.
3. Verify that the content fetching mechanism utilizes the configured gateway URLs.
4. Check the metrics to ensure the count of configured gateways is accurate.

### DIFF
--- a/packages/admin-ui/src/__mock-backend__/index.ts
+++ b/packages/admin-ui/src/__mock-backend__/index.ts
@@ -225,6 +225,7 @@ export const otherCalls: Omit<Routes, keyof typeof namedSpacedCalls> = {
   ipfsGatewayUrlsGet: async () => ({
     ipfsGatewayUrls: [
       "https://ipfs-gateway.dappnode.net",
+      "https://ipfs-gateway-dev.dappnode.net",
       "https://gateway.pinata.cloud",
       "https://cloudflare-ipfs.com"
     ]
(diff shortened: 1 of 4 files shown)


packages/dappmanager/src/calls/ipfsGatewayUrlsSet.ts
```diff
@@ -0,0 +1,41 @@
...
+    // Set new values in db
+    const gatewayUrls = nextGatewayUrls.length > 0 ? nextGatewayUrls : params.IPFS_REMOTE_URLS;
+    db.ipfsGatewayUrls.set(gatewayUrls);
+
+    // Change IPFS gateway URLs in the installer
+    dappnodeInstaller.changeIpfsGatewayUrls(db.ipfsGatewayUrls.get());
+  } catch (e) {
+    throw Error(`Error changing IPFS gateway URLs, ${e}`);
+  }
+}
```
Marketen: seems like `changeIpfsGatewayUrls()` and `ipfsGatewayUrlsSet()` are overlapping, they do the same or have too similar names

PR #2337, opened 2026-01-07
Implement UI activity tracking and metrics reporting

pablomendezroyo:
## Context

> This pull request introduces a new feature for tracking user activity within the UI. It reports metrics such as user activity status and session duration to the backend for monitoring and analysis.

## Approach

> The solution implements a custom hook to track user interactions and periodically report activity metrics to the backend. It includes metrics for user activity status, last activity timestamp, and session uptime, which are collected for Prometheus monitoring.

## Test instructions

1. Navigate to the application and interact with the UI (e.g., click, scroll, type).
2. Monitor the backend to ensure that user activity metrics are being reported correctly.
3. Verify that the metrics reflect the correct activity status and timestamps.
4. Test the application under different user interaction scenarios to ensure accurate reporting.


Pol-Lanski:
It is important that at the Ethical Metrics side we can know the history of an anonymous user. I assume we can do this with the TOR ID. The goal is to know when any given dappnode was connected and what they did. This way we can create categories according to behaviour:
- Those who only connect on the weekends
- Those who connect daily
- Those who connect weekly
- Those who connect monthly
And we will have a better profile of the users... without knowing anything about the users apart from that they exist!
Other common metrics that we will be able to extract with this are DAU and MAU (Daily and Monthly Active Users). 

Extra metrics that are interesting are also:
- Page views (which pages did the user visit? Settings? Stakers UI? Gnosis UI? Dappmanager Logs? VPN?)
- Method via which they connect: Tailscale, Wireguard, OpenVPN
- Packages installed

PR #2339, opened 2026-01-12
Updating Backup node FR to v2

mateumiralles:
Updating Backup node frontend according to the Backup node for validators v2 update. Now Supports mainnet, gnosis and hoodi, and each network has its own tab to activate/ deactivate the service


packages/types/src/routes.ts
```diff
@@ -656,13 +657,13 @@ export interface Routes {
    * Activates the beacon node backup
    * @param id the hashed license
    */
-  premiumBeaconBackupActivate: (id: string) => Promise<void>;
+  premiumBeaconBackupActivate: ({ key, network }: { key: string; network: Network }) => Promise<void>;
```
pablomendezroyo: it could be interesting to also add this as a new type, same way you did for `BeaconBackupNetworkStatus` and then use it everywhere instead of creating the type always

pablomendezroyo (changes requested):

PR #2340, opened 2026-01-13
remove backup env in stakers

mateumiralles:
remove backup env in stakers

### DIFF
--- a/packages/stakers/src/consensus.ts
+++ b/packages/stakers/src/consensus.ts
@@ -10,7 +10,8 @@ import {
   StarknetConsensusSepolia,
   Network,
   StakerItem,
-  UserSettings
+  UserSettings,
+  DetailsResponse
 } from "@dappnode/types";
 import { StakerComponent } from "./stakerComponent.js";
 import { DappnodeInstaller } from "@dappnode/installer";
@@ -266,7 +267,9 @@ export class Consensus extends StakerComponent {
   /**
    * Returns backup beacon node URL if premium license is active and valid, otherwise null
    */
-  private async getBackupIfActiveNoThrow(network: Network.Hoodi | Network.Mainnet | Network.Gnosis): Promise<string | null> {
+  private async getBackupIfActiveNoThrow(
+    network: Network.Hoodi | Network.Mainnet | Network.Gnosis
+  ): Promise<string | null> {
     try {
       const licenseRes = await fetch("http://premium.dappnode:8080/api/license");
       if (!licenseRes.ok) return null;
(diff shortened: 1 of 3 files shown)


packages/stakers/src/consensus.ts
```diff
@@ -170,6 +170,29 @@ export class Consensus extends StakerComponent {
...
+        if (isPrevInstalled) {
+          const composeEditor = new ComposeFileEditor(prevConsClientDnpName, false);
+          const validatorService = composeEditor.services()["validator"];
+          if (validatorService) {
+            validatorService.mergeEnvs({ BACKUP_BEACON_NODES: "" });
+            composeEditor.write();
+          }
+        }
+      } catch (e) {
+        console.error(`Failed to clear BACKUP_BEACON_NODES on previous client ${prevConsClientDnpName}:`, e);
```
pablomendezroyo: use internal module `logs` for consistance

packages/stakers/src/consensus.ts
```diff
@@ -185,10 +208,11 @@ export class Consensus extends StakerComponent {
     const isPkgInstalled = await this.isPackageInstalled(newConsensusDnpName);
     let environment = isPkgInstalled ? {} : getDefaultConsensusUserSettings({ network }).environment;
 
-    // Only for Mainnet and Hoodi, try to get backup beacon node
-    if (network === Network.Mainnet || network === Network.Hoodi) {
+    // Only for Mainnet, Gnosis and Hoodi, try to get backup beacon node
+    if (network === Network.Mainnet || network === Network.Gnosis || network === Network.Hoodi) {
       const backupUrl = await this.getBackupIfActive(network);
       if (backupUrl) {
+        console.log("BACKUP ACTIVE, Setting BACKUP_BEACON_NODES for", newConsensusDnpName, "to", backupUrl);
```
pablomendezroyo: module `logs`

packages/stakers/src/consensus.ts
```diff
@@ -185,10 +208,11 @@ export class Consensus extends StakerComponent {
     const isPkgInstalled = await this.isPackageInstalled(newConsensusDnpName);
     let environment = isPkgInstalled ? {} : getDefaultConsensusUserSettings({ network }).environment;
 
-    // Only for Mainnet and Hoodi, try to get backup beacon node
-    if (network === Network.Mainnet || network === Network.Hoodi) {
+    // Only for Mainnet, Gnosis and Hoodi, try to get backup beacon node
+    if (network === Network.Mainnet || network === Network.Gnosis || network === Network.Hoodi) {
       const backupUrl = await this.getBackupIfActive(network);
```
pablomendezroyo: if `getBackupIfActive` fails then `getUserSettings` will fail too and all the way up to `setStakerConfig` making this critical function call fail. Do we want to throw error and do not proceed with setting staker configuration if something wrong happen when trying to get if the backup was active?

Marketen: `getBackupIfActive` cant fail, it has everything inside a try catch that returns `null` if there were any errors

this hasnt changed in this PR

packages/stakers/src/consensus.ts
```diff
@@ -170,6 +170,29 @@ export class Consensus extends StakerComponent {
     });
     // persist on db
     if (newConsensusDnpName !== prevConsClientDnpName) await this.DbHandlers[network].set(newConsensusDnpName);
+
+    // If changing consensus clients and in a network where premium backup exists,
+    // always try to clear BACKUP_BEACON_NODES from the previous client
+    // Do this after everything else and only log error in console as this is not critical
+    if (
```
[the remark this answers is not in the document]
Marketen: done

PR #2455, opened 2026-07-23
fix: route avatars through configured IPFS gateways

3alpha:
## Context

Package avatar URLs were pinned to one public IPFS gateway in the browser, bypassing the configured local or remote gateway selection. Installed avatars also used a stable local URL even when a package changed its avatar, allowing the one-day browser cache to show stale content.

## Approach

- Route IPFS avatar URLs through an authenticated DAppManager endpoint keyed by the avatar CID.
- Refresh the installer gateway list from the current local/remote setting before retrieval and retain the existing multi-gateway fallback behavior.
- Validate CIDs, bound CAR and decoded avatar sizes, require a PNG signature, and return CID-addressed immutable responses.
- Add an 8 MiB byte-bounded memory LRU with in-flight request deduplication.
- Version installed local avatar URLs with the current IPFS or mirror avatar reference so package avatar changes invalidate browser caches.
- Prefer the current installed avatar URL in the public package manifest while retaining the persisted manifest fallback.

## Test instructions

1. Configure remote IPFS mode with multiple gateways, making the first gateway unavailable and leaving a later gateway healthy.
2. Open the package directory and verify IPFS avatars load through /avatar/ipfs/{cid}.
3. Switch to local IPFS mode and verify avatars resolve through the local gateway without changing Admin UI code.
4. Install or update a package whose avatar changes and verify its local URL keeps /avatars/{package}.png but receives a new v query value and displays the new image immediately.
5. Verify mirror-backed package avatars continue to load and that an updated mirror package changes the local avatar v query.
6. Run the focused avatar tests in utils, dockerApi, and dappmanager, then build the six affected TypeScript workspaces.

Automated validation completed: focused avatar tests pass; existing container parsing expectations pass; affected workspaces compile; changed files pass ESLint, Prettier, and git diff --check.

### DIFF
--- a/packages/admin-ui/server-mock/index.ts
+++ b/packages/admin-ui/server-mock/index.ts
@@ -69,5 +69,6 @@ startHttpApi({
   subscriptionsLogger,
   eventBus,
   isNewDappmanagerVersion: () => false,
+  getIpfsFileBytes: async () => new Uint8Array(),
   adminPasswordDb: new AdminPasswordDb(params)
 });
--- a/packages/dappmanager/src/api/avatar.ts
+++ b/packages/dappmanager/src/api/avatar.ts
@@ -1,0 +1,90 @@
+import { RequestHandler } from "express";
+import { CID } from "multiformats/cid";
+import { releaseFiles } from "@dappnode/types";
+
+const AVATAR_CACHE_MAX_BYTES = 8 * 1024 * 1024;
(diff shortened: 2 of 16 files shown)


hcastc00:
Fixed the gateway failover issue. Avatar retrieval now applies the existing 30-second IPFS timeout to each gateway attempt, allowing a stalled gateway to be aborted before trying the next configured gateway. Caller abort signals remain respected, and a regression test covers fallback from a hanging first gateway.

hcastc00 (approved):

PR #2459, opened 2026-08-25
Add a private mode switch to the Nexus chat

Marketen:
## Why

The chat talked to the gateway directly, which is the path where TLS terminates at Cloudflare and prompts are readable there. A toggle in the Nexus settings now routes it through `nexus-local-proxy` on this DAppNode, which verifies the gateway's AWS Nitro attestation against a pinned policy and encrypts request and response bodies with EHBP past that point.

Nothing else in the client had to change: the proxy is OpenAI-compatible on both endpoints this code uses — `/chat/completions` over the attested channel and `/models` passed through.

## Design notes

- **Persisted in the DB** alongside the API key rather than as an env var, so it survives restarts and is changeable from the UI.
- **Off by default.** The proxy fails closed, so opting in should be deliberate. The UI copy says so rather than presenting it as free, and links to the verification page.
- **An explicit `NEXUS_GATEWAY_URL` still wins** over the toggle, so a developer pointed at a staging gateway is not silently redirected to the local proxy. Covered by a test.

## Tests

13 passing. New coverage for both routing directions, the override precedence, persistence, and rejecting a non-boolean.

The one pre-existing test that asserted the whole status object needed the two new fields; its harness gained a private-mode store and a way to omit the gateway override.

`db`, `nexus`, `dappmanager` and `admin-ui` all typecheck clean.

## Note

This does **not** add a package dependency on `nexus-local-proxy` — it has no registry entry yet (`NOREPO`). The toggle works whenever the proxy is installed; it just does not assume it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/admin-ui/src/pages/nexus/api.ts
+++ b/packages/admin-ui/src/pages/nexus/api.ts
@@ -19,10 +19,15 @@ export interface NexusStatus {
   /** Whether the active key was pasted manually, created through Nexus login, or is unset. */
   keySource: "manual" | "nexus" | "none";
   accountLabel: string | null;
+  /** True when traffic is routed through the attested local proxy. */
+  privateMode: boolean;
+  /** Where the operator can inspect the attestation evidence. */
+  verificationUrl: string;
 }
 
 const STATUS_URL = "/nexus/status";
 const CONFIG_URL = "/nexus/config";
+const PRIVATE_MODE_URL = "/nexus/private-mode";
 const MODELS_URL = "/nexus/models";
 const CHAT_URL = "/nexus/chat/completions";
 const CONFIRM_URL = "/nexus/chat/confirm";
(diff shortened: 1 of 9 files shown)


Marketen:
Pushed a follow-up commit to this branch: **operator feedback on whether the local proxy can actually serve private mode.**

### Why

Private mode fails closed, so the first sign that `nexus-local-proxy` isn't installed was a chat message that wouldn't send, surfaced as an upstream error. The admin UI is exactly where someone who wouldn't think to check is most likely to flip the switch.

### What

`probeLocalProxy()` reads the proxy's own `/v1/verification` rather than attesting anything in dappmanager — one verifier on the node, everything else reads its answer. New `GET /nexus/private-mode/probe`, admin-only.

The key editor probes on open and again after a toggle, distinguishing the three cases that call for different action:

- **Not installed** → tells the operator to install the Nexus Local Proxy package
- **Installed but unverified** → says so, and that it won't carry prompts in that state
- **Verified** → names the attested release and check count, so the claim is visible rather than asserted

### Also

Corrected a comment describing the proxy as verifying "against a pinned policy." As of [dappnode-nexus-sdk#8](https://github.com/dappnode/dappnode-nexus-sdk/pull/8) it derives measurements from cosign-signed Gateway releases, so clients no longer carry pinned PCRs at all — which is also why this branch no longer needs updating when a Gateway release ships.

### Testing

4 new unit tests (verified / unreachable / running-unverified / unreadable response) — **17 passing**. `tsc --noEmit` clean on `nexus`, `dappmanager`, and `admin-ui`; eslint and prettier clean.

The branch merges cleanly on `develop` (3 commits behind, no conflicts).

